PR #40, opened 2021-04-20
adds endorsement table

sparkplug0025:


contracts/eden/src/inductions.cpp
```diff
@@ -15,12 +15,35 @@ namespace eden
          row.id = id;
          row.inviter = inviter;
          row.invitee = invitee;
-         row.witnesses = witnesses;
-         row.endorsements = {};
+         row.endorsements = witnesses.size() + 1;
```
swatanabe, line 18: `update_profile` also needs to be updated.

swatanabe (changes requested):

contracts/eden/src/inductions.cpp
sparkplug0025: nice catch! working on it!

sparkplug0025: done!

contracts/eden/src/inductions.cpp
```diff
@@ -29,10 +52,20 @@ namespace eden
...
-         row.new_member_profile = new_member_profile;
-         row.endorsements = {};
-      });
+      induction_tb.modify(induction_tb.iterator_to(induction), eosio::same_payer,
+                          [&](auto& row) { row.new_member_profile = new_member_profile; });
+
+      reset_endorsements(induction.id);
+   }
+
+   void inductions::reset_endorsements(uint64_t induction_id)
```
swatanabe, line 61: I'm assuming that `endorsements` is supposed to be the number of endorsements remaining and that it will be decremented with each endorsement.  If that's true, then it needs to be reset as well.

swatanabe (approved):

contracts/eden/src/inductions.cpp
sparkplug0025: (no, my idea was to have `endorsements` to be a fixed number of how many endorsements this induction process needs)

PR #45, opened 2021-04-21
adds genesis sample data

sparkplug0025:
with this we are able to see some induction sample data -- and adds a basic skeleton for the genesis process

### DIFF
--- a/contracts/eden/src/actions/genesis.cpp
+++ b/contracts/eden/src/actions/genesis.cpp
@@ -27,11 +27,10 @@ namespace eden
 
          for (const auto& endorser : genesis_pool)
          {
-            if (endorser == invitee)
+            if (endorser != invitee)
             {
-               continue;  // skips itself
+               inductions.create_endorsement(inviter, invitee, endorser, induction_id);
             }
-            inductions.create_endorsement(inviter, invitee, endorser, induction_id);
          }
       }
    }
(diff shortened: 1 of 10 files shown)


contracts/eden/src/actions/genesis.cpp
```diff
@@ -0,0 +1,43 @@
...
+      {
+         auto induction_id = i + 1;
+         const auto& invitee = genesis_pool[i];
+         auto total_endorsements = genesis_pool.size() - 1;
+
+         members.create(invitee);
+         inductions.create_induction(induction_id, get_self(), genesis_pool[i], total_endorsements);
+
+         auto endorsement = 0;
+         while (endorsement <= total_endorsements)
```
swatanabe, line 29: A range for loop over genesis_pool would be simpler.

contracts/eden/src/actions/genesis.cpp
```diff
@@ -0,0 +1,39 @@
...
+         members.create(invitee);
+         inductions.create_induction(induction_id, get_self(), genesis_pool[i], total_endorsements);
+
+         for (const auto& endorser : genesis_pool)
+         {
+            if (endorser == invitee)
+            {
+               continue;  // skips itself
+            }
+            inductions.create_endorsement(inviter, invitee, endorser, induction_id);
```
swatanabe, line 34: You could put this inside the if by flipping the condition and avoid the unstructured control flow.

swatanabe (approved):

PR #50, opened 2021-04-21
Convert tables to use variant

swatanabe:
Another solution which has the same binary layout, but requires more code changes when upgrading is to add a version byte which is always 0 as the first member of the struct.


tbfleming (approved):
Needs to run through clang-format. Approved.

tbfleming:
```
2021-04-26T14:23:33.6932745Z CMake Error at boost/CMakeLists.txt:15 (message):
2021-04-26T14:23:33.6933706Z   Download
2021-04-26T14:23:33.6934646Z   https://dl.bintray.com/boostorg/release/1.75.0/source/boost_1_75_0.tar.gz
2021-04-26T14:23:33.6935691Z   failed.  "HTTP response code said error"
```

This keeps going down sporadically. We'll probably have to set up a mirror.

PR #56, opened 2021-04-22
Implement genesis configuration.

swatanabe:


contracts/eden/src/actions/genesis.cpp
```diff
@@ -1,13 +1,32 @@
...
+                      eosio::asset minimum_donation,
+                      std::vector<eosio::name> initial_members,
+                      std::string genesis_video,
+                      eosio::asset auction_starting_bid,
+                      uint32_t auction_duration,
+                      eosio::ignore<std::string> memo)
    {
       require_auth(get_self());
 
+      eosio::check(community_symbol == minimum_donation.symbol, "community symbol does not match minimum donation");
```
sparkplug0025, line 19: please add a check for auction_starting_bid symbol as well

sparkplug0025 (approved):
lgtm; just one small check before merge

sparkplug0025:
I also just realized that your files are not being formatted with the config setup on `.clang-format`. Can you please apply the formatting before merge?

contracts/eden/src/actions/genesis.cpp
```diff
@@ -16,22 +35,24 @@ namespace eden
...
-         for (const auto& endorser : genesis_pool)
+         for (const auto& endorser : initial_members)
          {
             if (endorser != invitee)
             {
                inductions.create_endorsement(inviter, invitee, endorser, induction_id);
             }
          }
+
+         inductions.update_video(inductions.get_induction(induction_id), genesis_video);
```
sparkplug0025, line 55: I think it would be optimal if we add a new video string parameter (default to "") to `inductions.create_induction` method, then you can do it directly on emplace.

sparkplug0025: I actually was getting trx timeout with 3 members on genesis... not sure if it's because its always resetting the endorsements... (pretty small actually, 3*3)... I did this adjust in my branch and it passes (even with 20 members): https://github.com/eoscommunity/Eden/pull/57/files

swatanabe:
I really hate clang-format

PR #61, opened 2021-04-24
Filter notifications

tbfleming:


libraries/eosiolib/contracts/eosio/dispatcher.hpp
```diff
@@ -82,13 +84,13 @@ namespace eosio
...
-#define EOSIO_EXTRACT_IMPL1(x, n, r) r
-#define EOSIO_EXTRACT_IMPL2(r) EOSIO_EXTRACT_IMPL1(r)
-#define EOSIO_EXTRACT(base, x) EOSIO_EXTRACT_IMPL2(BOOST_PP_CAT(base, x))
 
 #define EOSIO_MATCH_NOTIFY(x) EOSIO_MATCH(EOSIO_MATCH_NOTIFY_, x)
-#define EOSIO_EXTRACT_NOTIFY(x) EOSIO_EXTRACT(EOSIO_MATCH_NOTIFY_, x)
 #define EOSIO_MATCH_NOTIFY_notify EOSIO_MATCH_YES
+#define EOSIO_EXTRACT_NOTIFY_CODE(x) BOOST_PP_CAT(EOSIO_EXTRACT_NOTIFY_CODE_, x)
+#define EOSIO_EXTRACT_NOTIFY_CODE_notify(code, action) code
+#define EOSIO_EXTRACT_NOTIFY_ACTION(x) BOOST_PP_CAT(EOSIO_EXTRACT_NOTIFY_ACTION_, x)
```
swatanabe, line 92: It's safe here because action names cannot contain an `_`, but as a general rule, this kind of concatenation should not use a trailing `_`, because if `x` begins with an `_`, it will create a reserved identifier.

PR #64, opened 2021-04-25
Ignores transfers from Eden Contract

sparkplug0025:


tbfleming:
It has a redundant check

PR #67, opened 2021-04-26
UI Tweaks and Responsive Work

brandonfancher:
I'm keeping it very high-level and wireframey right now. We can add color and more design later.

* Simplifies styling of navigation bar. Improves responsiveness.
* More coming...

### DIFF
--- a/packages/webapp/src/_app/ui/header-nav.tsx
+++ b/packages/webapp/src/_app/ui/header-nav.tsx
@@ -53,8 +53,7 @@ const HeaderItems = ({ menuItems }: { menuItems: MenuItem[] }) => {
     );
 };
 
-const HEADER_LINK_BASE_CLASS =
-    "mr-5 last:mr-0 hover:text-gray-900 ";
+const HEADER_LINK_BASE_CLASS = "mr-5 last:mr-0 hover:text-gray-900 ";
 
 const HeaderItemLink = ({
     href,
@@ -64,7 +63,8 @@ const HeaderItemLink = ({
 }: any) => {
     const { pathname } = useRouter();
     const active = exactPath ? pathname === href : pathname.startsWith(href);
-    const linkClass = HEADER_LINK_BASE_CLASS + (active ? "text-gray-900 font-bold" : "");
+    const linkClass =
+        HEADER_LINK_BASE_CLASS + (active ? "text-gray-900 font-bold" : "");
     return (
         <Link href={href}>
             <a className={linkClass} onClick={handleNavigationClick}>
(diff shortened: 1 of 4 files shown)


brandonfancher:
@sparkplug0025 Just some thoughts. The last link in this new nav bar style is "Membership." I originally had is set up so that it said "Join" if the user isn't logged in or is not a member of the Eden community and "Invite" if the user _is_ a member. But It resulted in a lot of flicker...because with every page switch, it would refetch that status beginning with undefined.

So, in the interest of keeping this PR scoped to UI, I'm not going to solve that right now... I just went with one word ("Membership") so there's no flicker.

That said, we are going to eventually want improvements, like displaying the user's profile image when logged in and their real name ("Brandon Fancher") instead of their EOS account name (if they're an Eden member.)

One way to accomplish these things is to introduce state so that profile information is only updated under certain conditions and there's no flicker. BUT...this is exactly the kind of thing React Query is for. So eventually (and not a huge priority now or even before we release) I'll layer in React Query.

brandonfancher:
I have some other UI work in flight. But I'm going to go ahead and PR this one.

packages/webapp/src/_app/ui/header-nav.tsx
```diff
@@ -13,66 +12,49 @@ interface MenuItem {
...
+                    {item.label}
+                </HeaderItemLink>
+            ))}
+        </>
     );
 };
 
 const HEADER_LINK_BASE_CLASS =
-    "block px-3 py-2 rounded-md text-base font-medium hover:bg-gray-700 hover:text-white ";
+    "mr-5 last:mr-0 hover:text-gray-900 ";
```
sparkplug0025, line 57: I feel like you need to install prettier and run it in the files that you edited. since the .prettierrc line lengh has enough space, this should be in a single line: 
`const HEADER_LINK_BASE_CLASS = "mr-5 last:mr-0 hover:text-gray-900 ";`

brandonfancher: Ah, possibly. Will do.

sparkplug0025 (approved):
I suspect you need to run prettier, else lgtm

sparkplug0025:
> One way to accomplish these things is to introduce state so that profile information is only updated under certain conditions and there's no flicker. BUT...this is exactly the kind of thing React Query is for. So eventually (and not a huge priority now or even before we release) I'll layer in React Query.

I feel like I need to tackle the pagination... if that's the case I will see how much overhead would be to implement by myself vs using ReactQuery

PR #70, opened 2021-04-26
api.ts: support variant rows

tbfleming:


contracts/eden/src/inductions.cpp
```diff
@@ -255,13 +255,13 @@ namespace eden
       {
          itr = endorsement_idx.erase(itr);
       }
-      induction_tb.erase(induction);
```
sparkplug0025, line 258: #69 might have a conflict because I'm fixing this right here to unblock our test environment

PR #73, opened 2021-04-27
deposit-and-spend

tbfleming:
Adopts deposit-and-spend pattern and moves the initiation fee to the final stage.

* The contract keeps track of deposits in the user's account
* The `inductpayfee` action spends the fee and mints NFTs
* The `withdraw` action withdraws funds from the user's account

Tests are in place for the deposit-and-spend accounting, but testing for the initiation flow is incomplete (todo).


contracts/eden/src/actions/induct.cpp
```diff
@@ -70,6 +73,21 @@ namespace eden
...
+   {
+      eosio::require_auth(payer);
+
+      globals globals{get_self()};
+      inductions inductions{get_self()};
+      accounts accounts{get_self()};
+
+      const auto& induction = inductions.get_induction(id);
+      eosio::check(payer == induction.invitee(), "only inductee may pay fee");
+      eosio::check(quantity == globals.get().minimum_donation, "incorrect fee");
```
sparkplug0025, line 86: I believe here we can verify for greater or equal than minimum donation fee (for the cases where the new member wants to donate a bigger amount)

contracts/eden/src/accounts.cpp
```diff
@@ -0,0 +1,47 @@
...
+      {
+         account_tb.modify(record, contract, [&](auto& a) { a.balance() += quantity; });
+      }
+   }
+
+   void accounts::sub_balance(eosio::name owner, const eosio::asset& quantity)
+   {
+      auto record = account_tb.find(owner.value);
+      eosio::check(record != account_tb.end() && record->balance() >= quantity,
+                   "insufficient deposit");
```
sparkplug0025, line 34: insufficient balance?

contracts/eden/src/accounts.cpp
```diff
@@ -0,0 +1,47 @@
...
+   }
+
+   void accounts::add_balance(eosio::name owner, const eosio::asset& quantity)
+   {
+      auto record = account_tb.find(owner.value);
+      if (record == account_tb.end())
+      {
+         // TODO: create another global
+         eosio::check(quantity >= globals.get().minimum_donation,
+                      "insufficient deposit to open an account");
```
sparkplug0025, line 20: I think we don't need to have a minimum deposit amount... eg: the user can deposit 5 and 5, and then complete his induction...

contracts/eden/include/accounts.hpp
```diff
@@ -0,0 +1,52 @@
...
+      {
+      }
+
+      std::optional<account> get_account(eosio::name owner);
+      void add_balance(eosio::name owner, const eosio::asset& quantity);
+      void sub_balance(eosio::name owner, const eosio::asset& quantity);
+
+      // this method is used only for administrative purposes,
+      // it should never be used outside genesis or test environments
+      void clear_all();
```
sparkplug0025, line 49: thanks for adding this. it helps a lot resetting the contract on testnets :)

contracts/eden/src/actions/induct.cpp
```diff
@@ -19,6 +20,8 @@ namespace eden
       {
          members.check_active_member(witness);
       }
+      if (members.is_new_member(invitee))
+         members.create(invitee);
       members.check_pending_member(invitee);
```
sparkplug0025, line 25: this `check_pending_member` can be inside an else branch because this check after creation would be redundant

contracts/eden/src/inductions.cpp
```diff
@@ -169,17 +167,15 @@ namespace eden
          endorsement_idx.modify(itr, eosio::same_payer, [](auto& row) { row.endorsed() = true; });
          itr++;
       }
-      maybe_create_nft(induction);
```
sparkplug0025, line 172: since you removed the `maybe_create_nft` from this function and we won't charge donation for genesis, you need to go inside the `inductprofil` and call `create_nft` after `endorse_all`. 

Now that I'm thinking about this function, I think we can rename this to `complete_genesis_induction` (since it's only used for this anyway) and instead of removing the `maybe_create_nft` just call `create_nft`.

sparkplug0025 (changes requested):
Looks great, thanks! I added some changes on my comments.

sparkplug0025:
Oh, just reminded about the abi too. Please add the new action to the abi.

contracts/eden/src/actions/induct.cpp
tbfleming: A separate donation action may help reduce mistakes. The UI could make it clear that donations aren't usable for induction or election fees.

contracts/eden/src/accounts.cpp
tbfleming: Since the contract pays for RAM on the row, we need some sort of mitigation against dust. This protection is defeatable by depositing 10.0000 and withdrawing 9.9999, but at least it protects against the common message deposit spam.

contracts/eden/src/inductions.cpp
tbfleming: I thought we were going to charge initiation fees for genesis. @brandonfancher? @cool-ant?

contracts/eden/src/accounts.cpp
tbfleming: changed

contracts/eden/src/actions/induct.cpp
tbfleming: changed

tbfleming:
abi updated

sparkplug0025 (approved):

PR #74, opened 2021-04-27
Create the collection and schema automatically on genesis.

swatanabe:


contracts/eden/src/atomicassets.cpp
```diff
@@ -51,4 +59,67 @@ namespace eden::atomicassets
...
+      else
+      {
+         actions::createcol(contract, {self, "active"_n})
+             .send(self, collection, true, std::vector{self}, std::vector{self}, 0.05,
+                   atomicassets::attribute_map{});
+      }
+
+      std::vector<format> schema{{"edenacc", "string"}, {"name", "string"},
+                                 {"img", "string"},     {"bio", "string"},
+                                 {"social", "string"},  {"inductionvid", "string"}};
```
sparkplug0025, line 92: Can we change from `edenacc` to `account` across the board? (to support multiple communities)

@brandonfancher I think that according to the best practices meeting we need these two changes:
1. rename `inductionvid` to `video`
2. new copyrights (or attributions) field? to follow the suggestion to credit the author of the picture art?

contracts/eden/src/actions/genesis.cpp
```diff
@@ -60,6 +61,9 @@ namespace eden
 
          induction_id++;
       }
+
+      atomicassets::init_collection(atomic_assets_account, get_self(), collection_name,
+                                    schema_name);
```
sparkplug0025, line 66: can we change the schema_name to `members` in our constants as well?

contracts/eden/src/atomicassets.cpp
```diff
@@ -51,4 +59,67 @@ namespace eden::atomicassets
...
+            // This will fail if the contract is not the author of the collection.
+            // In that case, it's the responsibility of the author to make sure
+            // that notifications are set up.
+            actions::addnotifyacc(contract, {self, "active"_n}).send(collection, self);
+         }
+      }
+      else
+      {
+         actions::createcol(contract, {self, "active"_n})
+             .send(self, collection, true, std::vector{self}, std::vector{self}, 0.05,
```
sparkplug0025, line 86: can we please move this magic 0.05 to a proper constant? (I believe it's the market fees)

contracts/eden/src/actions/genesis.cpp
sparkplug0025: nevermind, I just remembered that the schema name is set on compile time on `EDEN_SCHEMA_NAME`

contracts/eden/src/atomicassets.cpp
brandonfancher: Agreed. I'll create an issue for this. Thanks!

sparkplug0025 (approved):

PR #85, opened 2021-04-28
Invitation Tables & Related Layout/Responsive Updates

brandonfancher:
## What's In It?

This PR focuses on induction/invitation tables and container/card UI.
* Added responsive invitation tables for pending outgoing invites, incoming invites needing endorsement, incoming invites to invitees to join the community, etc.
* This wasn't quite possible without reworking some of the other container UI. I reworked the `RawLayout` and `SingleColLayout` a bit (hopefully simplifying) and made them more focused on layout and less opinionated about contents. I also introduced the concept of a `Card` that we put things in. At the `sm` size, it's 100% width. Padding is added as it scales up. 

## Fast follow
* I will need to polish some of the content above the invitation tables.
* Add boxes to a few other pages. 
* Adjust some wording here and there.

## Screenshots

image
image
image
image


packages/webapp/package.json
```diff
@@ -10,6 +10,7 @@
         "test": "yarn run build"
     },
     "dependencies": {
+        "dayjs": "^1.10.4",
```
sparkplug0025, line 13: dayjs <3 -- the days of moment are over 😢

packages/webapp/src/_app/layouts/card.tsx
```diff
@@ -0,0 +1,18 @@
+import { Heading } from "../ui";
+
+interface Props {
+    title?: string;
+    titleSize?: 1 | 2 | 3 | 4;
+    children: React.ReactNode;
+}
+
+export const Card = ({ children, title, titleSize = 1 }: Props) => (
```
sparkplug0025, line 9: When I think of Cards, I usually think about a component.... This `layouts` folder I created thinking about "page-layouts". I'd probably rename this Card to another "containerish/layouty" kinda name because I believe at some point we will want to create a generic card to use for posts, feeds etc. Unless you think this is the card component for it, so I'd probably move it to the ui components folder.

packages/webapp/src/inductions/api/eden-contract.ts
```diff
@@ -16,7 +16,8 @@ const INDEX_BY_ENDORSER = 2;
 const INDEX_BY_INDUCTION = 3;
 
 export const getInduction = async (
```
sparkplug0025, line 18: I love the way you decoupled the `getInduction` from `getEndorsementsByInductionId`! I'd just go further and not add the latest boolean flag on the `getInduction` function, we usually avoid adding boolean parameters to functions because it make it harder to read... eg: when you see this `getInduction(1234, true)` you don't know what this true stands for.

So I'd suggest us to go further and do this:
```ts
// function without boolean flag, only lookups for the induction
getInduction(1234)
// the new endorsements lookup: perfect
getEndorsementsByInductionId(1234)

// new function, this will call both above functions like the way you wrote with the boolean flag
getInductionWithEndorsements(1234) 
```

Another common approach is to have an optional object for options, if we think above function will grow and lookup for endorsements, another related members, or whatsoever:
```ts
getInduction(1234, { withEndorsements: true, withAnotherThing: true})
```

Ah, anyway, I love the refactor, these are just suggestions! :)

packages/webapp/src/inductions/components/induction-lists/action-button.tsx
```diff
@@ -0,0 +1,25 @@
+import NextLink from "next/link";
+
+interface Props {
+    href: string;
+    className?: string;
+    lightText?: boolean;
+    children: React.ReactNode;
+}
+
+export const InductionActionButton = ({
```
sparkplug0025, line 10: out of curiosity: any reason why we can't make this button a generic one inside _app/ui? (or even merge/mix with the current button?) Just wondering because I think we don't want to have different non-standard types of the buttons across the ui.

packages/webapp/src/inductions/components/induction-lists/induction-table.tsx
```diff
@@ -0,0 +1,136 @@
...
+interface InductionTableProps {
+    columns: Column[];
+    data: Row[];
+    className?: string;
+    headerClassName?: string;
+    rowClassName?: string;
+    tableHeader?: string;
+}
+
+export const Table = ({
```
sparkplug0025, line 29: Can we move this component to `_app/ui`? I'd love to reuse this nice table across the board :P

packages/webapp/src/inductions/components/induction-lists/invitee-inductions.tsx
```diff
@@ -0,0 +1,105 @@
...
+    {
+        key: "status",
+        label: "Action/Status",
+        type: InductionTable.DataTypeEnum.Action,
+    },
+];
+
+const getTableData = (inductions: Induction[]): InductionTable.Row[] => {
+    return inductions.map((ind) => {
+        const [allEndorsements] = useFetchedData<any>(
```
sparkplug0025, line 49: This is very interesting. I noticed that you did that in the above endorsements table too, but I never used a hook inside a function. I usually declare the hooks in the component and then I reuse the properties. Does it behave well with componentes re-rendering, updates and attributes changes? If yes, this is pretty cool! 🤯

packages/webapp/src/inductions/components/induction-lists/inviter-inductions.tsx
```diff
@@ -0,0 +1,134 @@
...
+        const [allEndorsements] = useFetchedData<any>(
+            getEndorsementsByInductionId,
+            ind.id
+        );
+        const endorsers = allEndorsements
+            ?.map((end: Endorsement): string => end.endorser)
+            .filter((end: string) => end !== ind.inviter)
+            ?.join(", ");
+        const remainingTime = dayjs().to(
+            dayjs(ind.created_at).add(7, "day"),
```
sparkplug0025, line 58: this is beautiful... `created_at` is an iso string and dayjs handles that out of the box? amazing!

packages/webapp/src/inductions/components/pending-inductions.tsx
```diff
@@ -22,27 +26,53 @@ export const PendingInductions = ({ isActive }: Props) => {
...
-                    </Text>
-                )}
-                <InductionsList inductions={inductions} isInviter={isActive} />
+                <>Loading inductions...</>
+            </div>
+        );
+    }
+
+    const thereAreEndorsements = !!userEndorsements.length;
+    const thereAreInductions = !!inductions.length;
```
sparkplug0025, line 42: nit: `x.length > 0` more readable than `!!`

sparkplug0025 (approved):
lgtm! That's a great PR!!! Nice improvements, good job man!!
I add some overall suggestions; feel free to implement it or discard if you think differently.

packages/webapp/package.json
brandonfancher: I know! Sad, right? Momentjs was my spacetime copilot for a long time!

packages/webapp/src/_app/layouts/card.tsx
brandonfancher: Yeah, I struggled a bit with where to put this one. I tried to keep the component super simple, because I do envision us eventually using this elsewhere, and multiple times per page. For example, on the members list page, it's used twice. I'll move it to `ui` for now. Thanks!

packages/webapp/src/inductions/api/eden-contract.ts
brandonfancher: Good suggestion. This was me trying to keep the PR changes minimal and adding new required attributes or changing the call signature too much meant touching additional files. But I think you're right; in this case, it's better to go ahead and do it. 👍

packages/webapp/src/inductions/components/induction-lists/action-button.tsx
brandonfancher: I wanted to keep this PR scoped as minimally as possible and tread lightly. But we're on the same wavelength...I'd like for this simpler button to potentially replace the other one across the UI. And I'm planning to experiment with that in a future PR. I'll keep this here for now...but you'll likely see this change happen very soon.

packages/webapp/src/inductions/components/induction-lists/induction-table.tsx
brandonfancher: Yeah, you know, I think maybe it's generic enough to be more than an _Induction_ table. I'll do a quick pass over it, remove the word Induction and Ind here and there, and move it to `_app/ui`. 👍

packages/webapp/src/inductions/components/induction-lists/invitee-inductions.tsx
brandonfancher: Yes, in this case it does behave reliably, but only because it's getting called in the same order with every render, it doesn't interact with state and the interplay with other hooks, _and_ I'm not calling it conditionally anywhere. 

That said, keeping hooks at the top level in a component _is_ a best practice, and best practices are best practices for a reason. If it's nested inside a function, I or others may be tempted to try to call that function conditionally in the future. Or it may later become dependent on other hooks in ways that complicate it.

I'm going to go ahead and move these to the top level. Good call.

packages/webapp/src/inductions/components/induction-lists/inviter-inductions.tsx
brandonfancher: Yeah, pretty cool, right!

packages/webapp/src/inductions/components/pending-inductions.tsx
brandonfancher: I can get on board with that! 😃

brandonfancher:
Thanks for the review. 😃  Comments addressed inline.

packages/webapp/src/inductions/components/induction-lists/invitee-inductions.tsx
brandonfancher: On second thought, I'm leaving this as is for now. Order of execution is guaranteed and these are simple implementations.

PR #93, opened 2021-04-30
clsdk is a package which supports building and testing contracts

tbfleming:


libraries/eosiolib/sdk/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    -lcltestlib
+)
+target_include_directories(cltestlib INTERFACE
+    catch2/include
+    contracts
+    eosiolib/contracts/include
+    eosiolib/tester/include
+)
+target_compile_options(cltestlib INTERFACE
+    -DCLSDK_CONTRACTS_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}/contracts/\"
```
swatanabe, line 95: I think it would be better if `set_code` had a search path for contracts.

libraries/eosiolib/sdk/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
+cmake_minimum_required(VERSION 3.16)
+project(clsdk)
```
swatanabe, line 2: Are users expected to pull this in using `add_subdirectory`?  For a pre-built package, support for `find_package`, i.e. clsdk-config.cmake, would be more idiomatic.

external/CMakeLists.txt
```diff
@@ -6,6 +6,13 @@ ProcessorCount(NUM_PROCS)
 add_library(rapidjson INTERFACE)
 target_include_directories(rapidjson INTERFACE rapidjson/include)
 
+if(IS_WASM)
+    copy_headers(rapidjson/include/rapidjson rapidjson/include/rapidjson)
+    copy_headers(rapidjson/include/rapidjson/error rapidjson/include/rapidjson/error)
+    copy_headers(rapidjson/include/rapidjson/internal rapidjson/include/rapidjson/internal)
```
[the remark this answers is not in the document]
tbfleming: This excludes the `msinttypes` dir

contracts/token/CMakeLists.txt
```diff
@@ -4,8 +4,15 @@ target_link_libraries(token eosio-contract-simple-malloc)
 set_target_properties(token PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${ROOT_BINARY_DIR})
 configure_file(token.abi ${ROOT_BINARY_DIR}/token.abi COPYONLY)
 
+configure_file(token.abi ${ROOT_BINARY_DIR}/clsdk/contracts/token.abi COPYONLY)
+configure_file(include/token/token.hpp ${ROOT_BINARY_DIR}/clsdk/contracts/token/token.hpp COPYONLY)
+add_custom_command(TARGET token POST_BUILD
+    COMMAND mkdir -p ${ROOT_BINARY_DIR}/clsdk/contracts
+    COMMAND cp -a $<TARGET_FILE:token> ${ROOT_BINARY_DIR}/clsdk/contracts
```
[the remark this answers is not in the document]
tbfleming: I could use it here, but it's pretty ornery for the more complex cases. cmake's replacements for unix file utils are incredibly poor.

libraries/eosiolib/sdk/clsdk-cmake-args
```diff
@@ -0,0 +1,4 @@
+SDK=$(realpath -e `dirname "$0"`/..)
+
+# User may override CMAKE_BUILD_TYPE by redefining it after `clsdk-cmake-args`
+echo -DCLSDK_PREFIX=$SDK -DCMAKE_TOOLCHAIN_FILE=$SDK/toolchain.cmake -DCMAKE_SYSROOT=${WASI_SDK_PREFIX}/share/wasi-sysroot -DCMAKE_BUILD_TYPE=Release
```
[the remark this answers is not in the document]
tbfleming: I wish. According to the docs, it appears so. When I tried it, it failed.

libraries/eosiolib/sdk/CMakeLists.txt
tbfleming: I did some experimenting and found a way to force the `find_package` approach to work by setting `clsdk_DIR`. `find_package`'s normal rules (e.g. searching through `CMAKE_PREFIX_PATH`) are borked by a poorly-documented and poorly-thought-out (IMO) feature: `find-package` prefixes paths in `CMAKE_PREFIX_PATH` with `CMAKE_SYSROOT`.

libraries/eosiolib/sdk/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    -Wl,--stack-first
+    -Wl,--entry,apply
+    -Wl,-z,stack-size=8192
+    -nostdlib
+    -leosio-contract-base
+)
+string(TOLOWER ${CMAKE_BUILD_TYPE} BUILD_TYPE_LOWER)
+if(BUILD_TYPE_LOWER STREQUAL release OR BUILD_TYPE_LOWER STREQUAL minsizerel)
+    target_link_options(eosio-contract-base INTERFACE -Wl,--strip-all)
+endif()
```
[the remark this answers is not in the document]
tbfleming: changed. cmake 3.16 needs `$<OR:...`

libraries/eosiolib/tester/eosio/tester.hpp
```diff
@@ -319,28 +318,37 @@ namespace eosio
          test_chain& t;
          std::vector<eosio::permission_level> level;
          std::optional<std::vector<std::vector<char>>> context_free_data;
+         std::optional<name> code;
 
          user_context with_cfd(std::vector<std::vector<char>> d)
          {
             user_context uc = *this;
             uc.context_free_data = std::move(d);
+            uc.level = {};
```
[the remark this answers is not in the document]
tbfleming: Yes. Fixed `act`.

libraries/eosiolib/sdk/CMakeLists.txt
tbfleming: I'm split. A search path would be convenient, but it could also lead to surprises.

swatanabe: A lot of tools have search paths, and I generally only see them causing surprises when they're overly complicated.  If we use a convention like `"clsdk/token.wasm"`, then there should be no ambiguity.

swatanabe (approved):

libraries/eosiolib/sdk/CMakeLists.txt
swatanabe: Compiling the SDK path into the test binary has the problem that it makes the test wasm completely non-portable.  For example, I'll run into problems if I try to compile in the cloud and download and run locally to debug failures.

tbfleming: Created #130

Issue #95, opened 2021-04-30
🎨 Adjust UI to include members full names instead of EOS account names where possible. Depends on #81 and deployment of related contract/table changes.


sparkplug0025:

sparkplug0025:
I think it's fully solved by #96 and #142

PR #96, opened 2021-04-30
member full names first stab

sparkplug0025:
first stab at #95 ; we need to revisit it when we implement react-query


packages/webapp/src/_app/eos/api.ts
```diff
@@ -14,7 +14,7 @@ export const getRow = async <T>(
     keyValue: string
 ): Promise<T | undefined> => {
     const rows = await getTableRows(table, keyValue);
-    return rows.length > 0 && `${rows[0][keyName]}` === keyValue
+    return rows.length > 0 && `${rows[0][keyName]}` === `${keyValue}`
```
sparkplug0025, line 17: this is the uint64 number/string bug fix

PR #106, opened 2021-04-30
Add cleanup of expired invitations and extra invitations.

swatanabe:
- Automatically clear up to 32 invitations for the inducted member when an induction is completed
- The remaining invitations and all expired invitations can be cleared by the gc action.


sparkplug0025:
I like the logic and the flow of the gc. Something that I'm not seeing is the removal of endorsements. Am I missing something?

sparkplug0025 (approved):
lgtm

PR #107, opened 2021-05-01
Introduce react-query with the current member name

brandonfancher:
## What is this?

This introduces the `react-query` package to the project and uses member names as an example. This particular PR showcases how much more efficient `react-query` can make our web app in terms of network request load.

Before, every link you clicked on would trigger another API call to lookup the name of the currently-logged-in user (and any other member names on the page on at a time) by their EOS account name. That's a lot of API calls for data we already have. Now (and without the introduction of a bunch of redux boilerplate) it looks it up once--when the user logs in or when a different user info is fetched for the first time. And when they sign out, the cache is cleared.

Also, when a non-member donates, the cache for their name is invalidated, which triggers a refetch to get their new full name from the member table and display it anywhere in the UI that's observing. (For example, after donation, their user name in the nav bar should automatically display.)

## Other uses for React Query

`react-query` will serve us well in other ways too--pagination comes to mind...or infinite scroll. And it really does replace the need much of the time for solutions like redux when dealing with fetched information. 

## Reading
* https://react-query.tanstack.com
* You should keep this handy at first too: https://react-query.tanstack.com/guides/important-defaults

## Follow-up work for another PR
* Anywhere we are displaying EOS account names but could instead display a member name, use this query!
* Other queries
* Pagination

### DIFF
--- a/packages/webapp/src/_app/ui/header-nav.tsx
+++ b/packages/webapp/src/_app/ui/header-nav.tsx
@@ -1,9 +1,9 @@
-import { getEdenMember } from "members";
 import Link from "next/link";
 import { useRouter } from "next/router";
-import { useEffect, useState } from "react";
+import { useQuery, useQueryClient } from "react-query";
 import { FaSignOutAlt } from "react-icons/fa";
 
+import { getEdenMember } from "members";
 import { useUALAccount } from "../eos";
 import { ActionButton } from "./action-button";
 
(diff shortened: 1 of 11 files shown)


brandonfancher:
Oh, I should mention the incredible dev tools too. It will show you the cache, the state of various queries, when things are fetching under the hood, etc. Great co-pilot. It can be accessed by clicking on the react-query icon bottom left of the web app window.

Once the React app is deployed in PROD mode, it is disabled.

Also, there's NextJS and SSR support...but I haven't played with that part of it yet.

packages/webapp/src/_app/ui/header-nav.tsx
```diff
@@ -77,31 +77,27 @@ const HeaderItemLink = ({
...
-        };
-        if (ualAccount && ualAccount.accountName) {
-            updateLoggedMemberName(ualAccount.accountName);
-        } else {
-            setMemberName("");
-        }
-    }, [ualAccount, accountName]);
+    const { data: member } = useQuery(
+        ["current-member", accountName],
+        async () => await getEdenMember(accountName),
```
brandonfancher, line 86: React Query actually automatically passes variables in to the function it calls. I'll do a slight refactor to allow for that: https://react-query.tanstack.com/guides/query-functions#query-function-variables

This will end up looking like this:

```js
const { data: member } = useQuery(
        ["current-member", accountName],
        getEdenMember,
        { staleTime: Infinity, enabled: !!ualAccount }
);
```

But I'll have to modify `getEdenMember` slightly.

packages/webapp/src/_app/ui/header-nav.tsx
```diff
@@ -78,30 +78,22 @@ const HeaderItemLink = ({
...
-            if (member) {
-                setMemberName(member.name);
-            }
-        };
-        if (ualAccount && ualAccount.accountName) {
-            updateLoggedMemberName(ualAccount.accountName);
-        } else {
-            setMemberName("");
-        }
-    }, [ualAccount, accountName]);
```
sparkplug0025, line 97: nice!

sparkplug0025 (approved):
awesome! 👏 👏 👏

PR #108, opened 2021-05-03
Sparkplug0025/UI errors

sparkplug0025:
It adds a proper toaster mechanism to throw user friendly errors.
image

It also adds some contract generic errors messaging.

### DIFF
--- a/contracts/eden/src/atomicassets.cpp
+++ b/contracts/eden/src/atomicassets.cpp
@@ -42,13 +42,11 @@ namespace eden::atomicassets
    {
       ::atomicassets::templates_t templates(contract, collection.value);
       const auto& templ = templates.get(
-          template_id,
-          std::string("fail to read template id " + std::to_string(template_id)).c_str());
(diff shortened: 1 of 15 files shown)


contracts/eden/src/atomicassets.cpp
```diff
@@ -41,9 +41,15 @@ namespace eden::atomicassets
                                      int32_t template_id)
    {
       ::atomicassets::templates_t templates(contract, collection.value);
-      const auto& templ = templates.get(template_id);
+      const auto& templ = templates.get(
+          template_id,
+          std::string("fail to read template id " + std::to_string(template_id)).c_str());
```
swatanabe, line 46: You don't need a cast.  It's already a string.

swatanabe (approved):

brandonfancher (approved):

contracts/eden/src/atomicassets.cpp
sparkplug0025: I cleaned them across the board

PR #110, opened 2021-05-03
Add test for post-genesis induction.

swatanabe:


contracts/eden/tests/test-eden.cpp
```diff
@@ -220,6 +257,70 @@ TEST_CASE("genesis")
...
+   // changing the video resets endorsements
+   t.pip.act<actions::inductvideo>("pip"_n, 4, bertie_video);
+   expect(t.bertie.trace<actions::inductdonate>("bertie"_n, 4, s2a("10.0000 EOS")),
+          "induction is not fully endorsed");
+   t.chain.start_block();
+   endorse_all();
+
+   t.bertie.act<actions::inductdonate>("bertie"_n, 4, s2a("10.0000 EOS"));
+   CHECK(get_eden_membership("bertie"_n).status() == eden::member_status::active_member);
+}
```
sparkplug0025, line 322: awesome tests! Can you add a few more? (it does not need to be this pr)
- trying to induct an existing member
- trying to induct with the inviter being repeated in the endorsement list
- trying to induct with repeated endorsers
- trying to induct with a pending member on the endorsement list
- trying to induct with a pending member as inviter
- complete an induction where the invitee has another X inductions and make sure the data was gc properly, when X has the following values:
  - X < 32
  - X = 32
  - X > 32

If it's too much we can split these in smaller tickets and I can tackle some when I finish the UI work.
cc @brandonfancher for visualization

sparkplug0025 (approved):

contracts/eden/tests/test-eden.cpp
swatanabe: > awesome tests! Can you add a few more? (it does not need to be this pr)
> ...
>     * trying to induct with a pending member as inviter

This one is already there.

PR #112, opened 2021-05-04
Responsive members grid and member holo card

brandonfancher:
This is a down payment / progress toward making member list grids responsive and styling the individual member pages. It introduces the `MemberHoloCard` component too...which is sort of a visual representation of the NFT that we'll be able to use elsewhere in the application.

Even though work is not complete on the individual member pages, I wanted to go ahead and PR this so other devs have the option to base on it if desired.

image
image
image
image


brandonfancher:
Images are a bit heavy. Will address that in subsequent PR.

packages/webapp/src/pages/members/[id].tsx
```diff
@@ -1,27 +1,40 @@
...
-            <MemberCard member={member} />
+        <RawLayout>
+            <Card>
+                <div className="flex justify-center items-center space-y-10 xl:space-y-0 xl:space-x-10 flex-col xl:flex-row">
+                    <div className="max-w-xl">
+                        <MemberHoloCard member={member} />
+                    </div>
+                    <MemberCard member={member} />
+                </div>
+            </Card>
```
Comment (sparkplug0025) [inline packages/webapp/src/pages/members/[id].tsx:26]: What if this entire <Card> is the <MemberCard>? I'm trying to think fo a case where we wnat the <MemberCard> without the HoloCard

packages/webapp/src/members/components/members-grid.tsx
```diff
@@ -1,17 +1,22 @@
 import React from "react";
 import Link from "next/link";
+import dayjs from "dayjs";
+import * as localizedFormat from "dayjs/plugin/localizedFormat";
 
 import { atomicAssets, ipfsBaseUrl } from "config";
 import { MemberData } from "../interfaces";
 import { assetToString } from "_app";
 
+dayjs.extend(localizedFormat.default);
```
sparkplug0025, line 10: is this dayjs supposed to exist in every file that we reference it or should we add this globally in our config.ts?

packages/webapp/src/members/components/members-grid.tsx
```diff
@@ -21,78 +26,124 @@ export const MembersGrid = ({ members }: Props) => {
...
+            <MemberImage member={member} />
+            <div className="p-3 pt-1">
+                <div className="flex items-center space-x-3 mt-1">
+                    <MintDate createdAt={member.createdAt} />
+                    <AssetBadge member={member} />
+                </div>
+                <MemberNames member={member} />
+                <AuctionBadge member={member} />
+                <SaleBadge member={member} />
+            </div>
```
sparkplug0025, line 43: beautiful refactor 💯

sparkplug0025 (approved):
approved! super beautiful improvements! 🥺

packages/webapp/src/pages/members/[id].tsx
Comment (brandonfancher) [reply packages/webapp/src/pages/members/[id].tsx:26]: I’m going to be doing some more work on this page today. I’ll def take a look at consolidating this. Thanks!

packages/webapp/src/members/components/members-grid.tsx
brandonfancher: I think you’re right; prob at the top level. I’ll play with it today.

packages/webapp/src/members/components/members-grid.tsx
brandonfancher: Thanks!

PR #114, opened 2021-05-04
Add Attributions on UI

sparkplug0025:
- add attributions field on the profile form
- add attributions label on member card preview
- add attributions on the member image title tag


packages/webapp/src/members/components/member-holo-card.tsx
```diff
@@ -24,6 +24,7 @@ export const MemberHoloCard = ({ member }: { member: MemberData }) => {
                 <img
                     src={`${ipfsBaseUrl}/${member.image}`}
                     className="rounded-full object-cover bg-white"
+                    title={member.attributions || undefined}
```
brandonfancher, line 27: So we don't leave this `undefined`, maybe: 

```
title={
    `Member image for ${member.name}.` +
    (member.attribution
        ? ` Attribution: ${member.attribution}`
        : "")
}
```

there's prob a better syntax for that

brandonfancher (approved):

Issue #117, opened 2021-05-04
Member profile Page Titles


sparkplug0025:
Go to a member profile page, eg https://eden-dev.vercel.app/members/egeon.edev

Current behavior:
- Html Page title is plain `Eden`

Expected behavior:
- Html Page title should be `Egeon's Profile | Eden`
- working reference: https://eden-prod.vercel.app/members/edenmember22

Issue #116, opened 2021-05-04
Sales badge is not working


sparkplug0025:
I prepared this scenario on egeon.edev: https://eden-dev.vercel.app/members/egeon.edev -- scroll down his collection until you see the member 33
![image](https://user-images.githubusercontent.com/79721020/117015304-95fc4100-acbf-11eb-97bc-9c540cc50a38.png)

Click in the ON SALE badge

Current behavior:
- Overrides the page with the eden member profile

Expected behavior:
- Open the atomic market link a new tab

Issue #115, opened 2021-05-04
Members cards are not being able to open in a new tab


sparkplug0025:
Hold CMD + click in a member card

current behavior:
- overriding current page

expected behavior:
- open the page in a new tab

PR #118, opened 2021-05-04
Rollup of misc fixes

brandonfancher:
This addresses #115, #116 and #117 and also consolidates `dayjs` setup in the `config.ts` file.

### DIFF
--- a/packages/webapp/src/config.ts
+++ b/packages/webapp/src/config.ts
@@ -1,3 +1,11 @@
+// dayjs setup
+import dayjs from "dayjs";
+import * as localizedFormat from "dayjs/plugin/localizedFormat";
+import * as relativeTime from "dayjs/plugin/relativeTime";
+dayjs.extend(localizedFormat.default);
+dayjs.extend(relativeTime.default);
+// end dayjs setup
+
 import { assetFromString } from "./_app/utils/asset";
 
 if (
--- a/packages/webapp/src/config.ts
+++ b/packages/webapp/src/config.ts
@@ -1,11 +1,3 @@
-// dayjs setup
-import dayjs from "dayjs";
-import * as localizedFormat from "dayjs/plugin/localizedFormat";
-import * as relativeTime from "dayjs/plugin/relativeTime";
-dayjs.extend(localizedFormat.default);
-dayjs.extend(relativeTime.default);
-// end dayjs setup
-
 import { assetFromString } from "./_app/utils/asset";
 
 if (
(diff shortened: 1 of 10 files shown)


packages/webapp/src/config.ts
```diff
@@ -1,3 +1,11 @@
+// dayjs setup
+import dayjs from "dayjs";
+import * as localizedFormat from "dayjs/plugin/localizedFormat";
+import * as relativeTime from "dayjs/plugin/relativeTime";
+dayjs.extend(localizedFormat.default);
+dayjs.extend(relativeTime.default);
+// end dayjs setup
```
sparkplug0025, line 7: I think I mentioned this on the config.ts file but since it's just app setup I wonder if it should stay at the `src/_app.tsx`. Thoughts?

packages/webapp/src/members/components/members-grid.tsx
```diff
@@ -1,18 +1,20 @@
...
-dayjs.extend(localizedFormat.default);
-
 interface Props {
     members: MemberData[];
 }
 
+const openInNewTab = (url: string) => {
+    const newWindow = window.open(url, "_blank", "noopener,noreferrer");
+    if (newWindow) newWindow.opener = null;
+};
```
sparkplug0025, line 16: oh wow, we really needed this? I wonder why the traditional `<a href={x} target="_blank">` does not work....

sparkplug0025 (approved):
lgtm! couple of comments/questions!

packages/webapp/src/config.ts
brandonfancher: That's where I went first...and then I read your original comment. haha! I agree with you; I'll put it in `_app.tsx`. 👍

packages/webapp/src/members/components/members-grid.tsx
brandonfancher: I tried that but them Chrome/Brave loudly reminded me that you can't have an `<a>` tag within an `<a>` tag. Not valid. So this was the only real way to get the Cmd-click behavior on the card body _and_ on the labels. At least that I could think of before I moved on. 🤷

brandonfancher: I also tried factoring away from an `<a>` tag at the card level, but again, not the behavior I wanted.

sparkplug0025: Ohhh I see... yeah the badges will overlap with the card footer <a>... :(

PR #120, opened 2021-05-04
Finish responsive work on member detail page

brandonfancher:
## What's in this PR?
* Finish responsive cleanup of member detail page
* Improve "Ready to join Eden?" page EOS users who aren't in Eden see when they want to join.
* Adds "Eden on EOS" logo to nav bar!!!

## Screenshots
image
image
image
image
image


packages/webapp/src/_app/ui/social-button.tsx
```diff
@@ -14,6 +15,7 @@ const BASE_CLASS = "hover:underline items-center flex space-x-2";
 export const SocialButton = ({
     handle,
     color,
+    size = 6,
```
brandonfancher, line 18: This is sort of a hack to work around the issue where passed-in`className`s don't override those in the component. Maybe we refactor this component in the future to be more like ActionButton without too many built-in classNames. Works for now though.

packages/webapp/src/inductions/components/induction-journey-container.tsx
```diff
@@ -0,0 +1,64 @@
+import React from "react";
```
brandonfancher, line 1: Prob not quite evident from this PR, but I'm planning to reuse this component in a few places in the induction process.

packages/webapp/src/members/components/member-card.tsx
```diff
@@ -13,22 +13,73 @@ interface Props {
...
+                    size={5}
+                    href={`${ipfsBaseUrl}/${member.inductionVideo}`}
+                />
             </div>
         </div>
     );
 };
+
+const MemberBio = ({ bio }: { bio: string }) => {
+    const [expanded, setExpanded] = useState(false);
```
brandonfancher, line 35: The UI doesn't quite look right when the bio is longer than 300 or so chars...it's a lot of text. This collapses it and makes it easy to toggle expanded state.

packages/webapp/src/_app/ui/social-button.tsx
sparkplug0025: yeah, I just created #124 to tackle these dynamic namings...

packages/webapp/src/members/components/member-card.tsx
sparkplug0025: now you are showing off

packages/webapp/src/members/components/member-collections.tsx
```diff
@@ -37,33 +37,27 @@ export const MemberCollections = ({ account, templateId }: Props) => {
...
+            </Button>
+            <Button
+                color="gray"
+                outline={tab !== "collectedBy"}
+                disabled={tab === "collectedBy"}
+                onClick={() => setTab("collectedBy")}
+                className="ml-4"
+            >
+                Collected By
+            </Button>
```
sparkplug0025, line 57: Although I implemented this as buttons, ideally it should be implemented with Tabs. It was just a rush attempt to build these collections cards.

packages/webapp/src/members/components/member-social-links.tsx
```diff
@@ -28,65 +28,68 @@ const urlify = (address: string) => {
...
-                    icon={FaFacebook}
-                    color="indigo"
-                    href={`https://facebook.com/${member.socialHandles.facebook}`}
-                />
-            )}
-        </div>
+        )}
+        {member.socialHandles.blog && (
+            <SocialButton
+                handle="Blog"
```
sparkplug0025, line 50: now that I'm thinking, I believe we should change this icon to a link icon (eg: 
image) and then replace the Blog word with the url... thoughts? not for this PR too.

sparkplug0025 (approved):
this is great!!! 👏

packages/webapp/src/members/components/member-collections.tsx
brandonfancher: Makes sense. I'm in "rush mode" too. Maybe after our first release, we can take a few days to clean up things like this.

packages/webapp/src/members/components/member-social-links.tsx
brandonfancher: Yeah, I agree.

PR #122, opened 2021-05-05
Page Transitioning Loading Bar Indicator

sparkplug0025:
Adds a transition loader when changing pages.

The loading bar is blue... I was about to change to orange but I realized in our UI/UX docs that blue seems to be our color, so I decided to not change it.


brandonfancher (approved):
😍

PR #125, opened 2021-05-05
Member stats should handle expired invitations

swatanabe:


sparkplug0025 (approved):
lgtm! nice catch on handling memberstats!

Issue #80, opened 2021-04-28
assess which wallets we want to support


sparkplug0025:
for now we support only Anchor...

Which ones should we support?
- Scatter
- Wombat
- Lynx
- etc..

brandonfancher:
One of the new `UALProvider` authenticators is erroring out. In dev mode, it's loud (see below). I don't think it impacts user experience in prod mode though. But it is annoying in dev mode.

![image](https://user-images.githubusercontent.com/7911424/117218774-cfbd6c80-add1-11eb-94c5-dcc91eaf0393.png)
![image](https://user-images.githubusercontent.com/7911424/117218789-d6e47a80-add1-11eb-804b-6323be6be56e.png)

I think it's Scatter. Not sure.

PR #126, opened 2021-05-05
adds scatter and ledger

sparkplug0025:
relates to #80 

Added wallets:
- Scatter
- Ledger

Tried to add these but I couldn't...
- Wombat: does not have UAL support
- Simpleos: does not have UAL support
- Trezor: does not have UAL support
- Lynx: their SSL certificates are expired, their links to create wallets are broken etc.


brandonfancher:
@sparkplug0025 Got my test Ledger fired up and key added to the `alice.edev` WAX Testnet account. Works beautifully!

brandonfancher (approved):

PR #145, opened 2021-05-06
Basic support for the election process

swatanabe:
Dividing members into groups
============================

The group size is 4-6, except for the last round

Randomizing the groups
======================

Voting
======

When a group has reached consensus, each member should report his
vote. 

Once the voting period closes, the election results will be processed.

If more than two thirds of the group members have voted for the
same candidate, the winner advances to the next round.
Otherwise, no one advances to the next round.

Eligibility to Vote
===================

A member is eligible to participate in an election if he donated 2 EOS and opted in to the election.

Members who were inducted at least 30 days before the start of the election, but did not donate 2 EOS will be kicked at the start of the election.

Election Schedule
=================

Elections are automatically scheduled on a configurable day of the week and time.
An election is automatically scheduled at the first such time at least 180 days
after the start of the previous election.  Several events can cause the election
to be moved to the first such time at least 30 days from the triggering event.

- The election schedule for 30 days from the current time is locked.  If there is an election, it cannot be moved.  If there is no election, then none can be scheduled.
- An election is always scheduled
- An election can only be moved sooner, never later.

Election Results
================

The lead representative and the board are tracked in a singleton
that can be checked in actions that require their authorization.

Actions
=======

```
void electdonate(eosio::name member, eosio::asset amount);
void electopt(eosio::name member, bool include_me);
void electseed(const eosio::bytes& btc_block_header);
void electprocess(uint32_t max_steps);
void electvote(uint8_t round, eosio::name voter, eosio::name candidate);
```


contracts/eden/include/eden.hpp
```diff
@@ -54,6 +55,9 @@ namespace eden
                    atomicassets::attribute_map collection_attributes,
                    eosio::asset auction_starting_bid,
                    uint32_t auction_duration,
+                   uint8_t election_day,
```
tbfleming, line 58: Please put the new fields at the end to aid history

tbfleming:
`elect.state` and `elect.curr` are missing in the ABI

tbfleming:
`bylaws`, `distaccount`, `distribution`, `pools`, and `votes` are also missing

tbfleming:
`clearall` doesn't clear the `bylaws` table

contracts/eden/src/actions/induct.cpp
```diff
@@ -98,7 +100,7 @@ namespace eden
       }
       else if (migrations.is_completed<migrate_account_v0>())
       {
-         internal_accounts.add_balance("master"_n, quantity);
+         add_to_pool(get_self(), "master"_n, quantity);
```
tbfleming, line 103: `internal_accounts` (above) is no longer used

contracts/eden/src/distributions.cpp
```diff
@@ -0,0 +1,363 @@
...
+      members members{contract};
+      distribution_account_table_type dist_accounts_tb{contract, default_scope};
+      auto dist_idx = dist_accounts_tb.get_index<"byowner"_n>();
+      auto dist_iter = dist_idx.find(distribution_account_key(contract, dist.distribution_time, 0));
+      auto& table = members.get_table();
+      auto iter = table.upper_bound(dist.last_processed.value);
+      auto end = table.end();
+      for (; max_steps > 0 && iter != end; ++iter, --max_steps)
+      {
+         auto member = *iter;
```
tbfleming, line 200: not used

contracts/eden/src/elections.cpp
```diff
@@ -0,0 +1,774 @@
...
+      eosio::check(election_time >= lock_time, "New election time is too close");
+      if (state_sing.exists())
+      {
+         auto state = state_sing.get();
+         eosio::check(
+             std::holds_alternative<current_election_state_registration>(state) &&
+                 std::get<current_election_state_registration>(state).start_time >= lock_time,
+             "Election cannot be rescheduled");
+      }
+      state_sing.set(current_election_state_registration{election_time}, contract);
```
tbfleming, line 253: Leaves `election_threshold` uninitialized

contracts/eden/src/elections.cpp
```diff
@@ -0,0 +1,774 @@
...
+         // Ignore events that would trigger an election unless they move
+         // the next election closer.
+         auto current_state = state_sing.get();
+         if (auto* current = std::get_if<current_election_state_registration>(&current_state))
+         {
+            auto new_start_time = eosio::block_timestamp{
+                get_election_time(state.election_start_time, now + eosio::days(30))};
+            if (new_start_time < current->start_time)
+            {
+               state_sing.set(current_election_state_registration{new_start_time}, contract);
```
tbfleming, line 321: Leaves `election_threshold` uninitialized

contracts/eden/src/elections.cpp
```diff
@@ -0,0 +1,774 @@
...
+      }
+
+      bylaws bylaws(contract);
+      bylaws.new_board();
+   }
+
+   uint32_t elections::randomize_voters(current_election_state_init_voters& state,
+                                        uint32_t max_steps)
+   {
+      election_state_singleton sequence_state(contract, default_scope);
```
tbfleming, line 388: not used

contracts/eden/src/elections.cpp
```diff
@@ -0,0 +1,774 @@
...
+            {
+               case no_donation:
+               {
+                  distributions distributions{contract};
+                  if (!state.last_closed_account)
+                  {
+                     state.last_closed_account = 0;
+                  }
+                  max_steps = distributions.on_election_kick(iter->account(),
+                                                             *state.last_closed_account, max_steps);
```
tbfleming, line 407: `on_election_kick` doesn't use its 2nd arg

contracts/eden/src/elections.cpp
swatanabe: It's actually 0.  It will won't cause any problems, but will cause a small amount of extra work every time a member is inducted.  I can set it to 10,001, to avoid this.

contracts/eden/src/actions/accounts.cpp
```diff
@@ -91,6 +95,50 @@ namespace eden
...
+      eosio::require_auth(from);
+      members members{get_self()};
+      members.check_active_member(from);
+      members.check_active_member(to);
+      distribution_account_table_type distribution_accounts{get_self(), default_scope};
+      accounts accounts{get_self()};
+      const auto& row = distribution_accounts.get_index<"byowner"_n>().get(
+          distribution_account_key(from, distribution_time, rank));
+      eosio::check(row.balance() >= amount, "insufficient balance");
+      distribution_accounts.modify(row, get_self(), [&](auto& row) { row.balance() -= amount; });
```
tbfleming, line 124: The row sticks around forever with an empty balance, unless `on_election_kick` or `on_resign` happens. Is this intended?

contracts/eden/src/elections.cpp
```diff
@@ -0,0 +1,770 @@
...
+   {
+      std::vector<eosio::name> result;
+      auto vote_idx = vote_tb.get_index<"bygroup"_n>();
+      for (auto iter = vote_idx.begin(), end = vote_idx.end(); iter != end;)
+      {
+         result.push_back(iter->member);
+         eosio::check(result.size() < 12, "Too many board members");
+         iter = vote_idx.erase(iter);
+      }
+      eosio::check(!result.empty(), "No board members");
```
tbfleming, line 551: This will hang the contract if no groups in the previous level reached consensus

swatanabe: Yep.  I don't know what we should do in this case.

tbfleming: Maybe create an issue to track this

tbfleming (approved):
Ready to merge. We probably need to track the above 2 things somewhere (e.g. issues).

tbfleming (approved):

PR #158, opened 2021-05-07
Init Cypress and adds a basic CI

sparkplug0025:
With this we have an initial version of Cypress E2E tests... Checkout how cool Cypress record the e2e pages and saves it in the build artifacts: eoscommunity/Eden/actions/runs/820293388

The video is actually not 100% good, but I think the best feature is when the test fails and we will have screenshots of the page.

I think this is the best ROI for tests... We will achieve a nice spot when we get these scenarios:
- Signin with an existing member and create a new invite
- Complete an induction (yes, signin with all of the members and complete the necessary steps for the induction)
- See the new member profile page
- See the new member in the new members section of community page

By having that we can rest assured that our changes in the contract and ui didn't affect the main induction flow. (In an automated way)

### DIFF
--- a/.github/workflows/webapp.yml
+++ b/.github/workflows/webapp.yml
@@ -20,6 +20,33 @@ on:
             - "yarn.lock"
 
 jobs:
+    webapp-e2e:
+        name: WebApp E2E Tests
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: ✅ Checkout code
+              uses: actions/checkout@v2
+
+            - name: 🛠 Build and Start WebApp
+              run: |
+                  yarn
+                  yarn build --stream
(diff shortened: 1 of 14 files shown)


sparkplug0025:
With this we have an initial version of Cypress E2E tests... Checkout how cool Cypress record the e2e pages and saves it in the build artifacts: https://github.com/eoscommunity/Eden/actions/runs/820293388

The video is actually not 100% good, but I think the best feature is when the test fails and we will have screenshots of the page.

brandonfancher:
This is AWESOME. Definitely the best bang for our buck in terms of testing. Good call, @sparkplug0025. And thanks for the orientation on zoom.

brandonfancher (approved):

Issue #103, opened 2021-04-30
🎨 Add site footer


brandonfancher:
Not sure what all we'll put in it for now. At least a link to this GH repo. And we can prob just use something from https://tailblocks.cc.

PR #164, opened 2021-05-10
issues/103: Add site footer

NorseGaud:
https://github.com/eoscommunity/Eden/issues/103


brandonfancher:
Nathan! Thanks for the PRs, man! This is a great first issue and something we definitely need. A few things I'd like to see before we merge this:

* It should stick to the bottom of the window.
image
* It should appear on every page. I noticed I don't see it on `/community` or `/members`
* I also recommend using the Eden SVG logo here: 
image
* Let's remove the Copyright for now (and 2020). We'll find something else to put down there soon. Our copyright is the right to copy. haha.

You'll prob want to start with a rebase.

Thanks!

NorseGaud:
@brandonfancher got it! Tailwind is new for me and I'm liking it :)

> I also recommend using the Eden SVG logo here:

So, this looks really bad on the first page since the logo shows twice in the same side, on the same page. I opted to just add "We are..." which flows into the statement.

brandonfancher (changes requested):
Yeah, this is coming together! As can be seen in the screenshot below, for some reason, it's making the document larger than the viewport height by a few pixels. It scrolls when it shouldn't. Is that happening for you? Maybe you can track that down?

image

After that's resolved, I'll massage some of the links in the footer and we'll get this merged.

NorseGaud:
@brandonfancher , good catch! It was a simple padding issue.

brandonfancher:
Almost there! (And sorry to keep sending feedback piecemeal.)

We need to ensure everything's responsive.

image

brandonfancher (changes requested):
See previous comment (responsive).

NorseGaud:
> See previous comment (responsive).

Could you clarify for me what specifically is wrong? Maybe detail what in the screenshot is wrong?

NorseGaud:
@brandonfancher I'm not seeing what you're seeing (I might not have access). If the footer doesn't fit with a page that I don't see, could we merge this and you add the small changes necessary to get it to play well? All of the pages I seem to have access to work fine.

Also, could you provide a little more specific information about the concerns? The screenshot helps, but I end up having to guess as to what you really want.

brandonfancher:
> @brandonfancher I'm not seeing what you're seeing (I might not have access). If the footer doesn't fit with a page that I don't see, could we merge this and you add the small changes necessary to get it to play well? All of the pages I seem to have access to work fine. Otherwise, how do I reproduce what you're seeing?

Regarding the responsive problems, I'm referring to the home page, not private pages that you wouldn't have access to.

> Also, could you provide a little more specific information about the concerns? The screenshot helps, but I end up having to guess as to what you really want.

If we just drop the links _under_ the "We are..." paragraph at smaller breakpoints and with appropriate padding, we should be good.

I see you just pushed another tweak. I'll take a look when I have a chance.

github-service-user-veertu:
> > @brandonfancher I'm not seeing what you're seeing (I might not have access). If the footer doesn't fit with a page that I don't see, could we merge this and you add the small changes necessary to get it to play well? All of the pages I seem to have access to work fine. Otherwise, how do I reproduce what you're seeing?
> 
> Regarding the responsive problems, I'm referring to the home page, not private pages that you wouldn't have access to.
> 
> > Also, could you provide a little more specific information about the concerns? The screenshot helps, but I end up having to guess as to what you really want.
> 
> If we just drop the links _under_ the "We are..." paragraph at smaller breakpoints and with appropriate padding, we should be good.
> 
> I see you just pushed another tweak. I'll take a look when I have a chance.

Ah, yes, I see, stacked links is fine and possible. I'll make a change tonight to do that. Thanks @brandonfancher

NorseGaud:
> > @brandonfancher I'm not seeing what you're seeing (I might not have access). If the footer doesn't fit with a page that I don't see, could we merge this and you add the small changes necessary to get it to play well? All of the pages I seem to have access to work fine. Otherwise, how do I reproduce what you're seeing?
> 
> Regarding the responsive problems, I'm referring to the home page, not private pages that you wouldn't have access to.
> 
> > Also, could you provide a little more specific information about the concerns? The screenshot helps, but I end up having to guess as to what you really want.
> 
> If we just drop the links _under_ the "We are..." paragraph at smaller breakpoints and with appropriate padding, we should be good.
> 
> I see you just pushed another tweak. I'll take a look when I have a chance.

Ah yes, I see. Stacking is possible and I'll work on that tonight. Thanks @brandonfancher

NorseGaud:
@brandonfancher I think I got it this time ;)

brandonfancher (approved):
Looks great. 💥  I'll get this merged. As for the links, I'll massage those post merge. 

This is awesome. Excellent contribution. Thank you, @NorseGaud!

Issue #127, opened 2021-05-05
eden: modify blog link icon and text


brandonfancher:
now that I'm thinking, I believe we should change this icon to a link icon (eg: 
![image](https://user-images.githubusercontent.com/79721020/117133889-d5379a00-ad72-11eb-9558-352e10a62678.png)) and then replace the Blog word with the url... thoughts? not for this PR too.

_Originally posted by @sparkplug0025 in https://github.com/eoscommunity/Eden/pull/120#discussion_r626481574_

sparkplug0025:
@NorseGaud here: https://github.com/eoscommunity/Eden/blob/main/packages/webapp/src/members/components/member-social-links.tsx#L48-L56

NorseGaud:
Ah, an area I can't seem to access from the app even after logging in -- is that right?

sparkplug0025:
Actually you can. Just click in any member that has a blog, eg: https://eden-dev.vercel.app/members/test255.edev

![image](https://user-images.githubusercontent.com/79721020/117653667-b2d2c180-b162-11eb-8cf2-acf9874f05e2.png)

See the blog link? it should be changed to the url icon and have the domain displayed there instead.

NorseGaud:
Oh wow, I totally missed that. Thanks!

brandonfancher:
Icon replaced (thanks @NorseGaud). We still need the word "Blog" replaced with the URL. Keeping this issue open.

sparkplug0025:
@brandonfancher I think we should have some formatting here just to be beautiful displayed...

Let's say we have the following scenarios:
- moreequalanimals.com
- https://moreequalanimals.com
- https://moreequalanimals.com/this-is-a-great-article-written-by-daniel-larimer-about-xyz

I think for the first 2 cases we should display the domain (not the url), so: moreequalanimals.com

For the last one, it would be a long string and probably break the ui. So I believe something like `moreequalanimals.com/this-is-a-great...` would be proper.

Any thoughts?

NorseGaud:
> For the last one, it would be a long string and probably break the ui. So I believe something like `moreequalanimals.com/this-is-a-great...` would be proper.

Could you clarify what "break the ui" means? What would break? (I'm partially asking because I hate when sites shorten URLs like that :) )

brandonfancher:
Yeah, good point. I'm fine with this proposal...drop the protocol and truncate with ellipsis after certain length. 

[...]

PR #168, opened 2021-05-11
issues/127: replaced blog icon with link

NorseGaud:
https://github.com/eoscommunity/Eden/issues/127


sparkplug0025 (approved):
lgtm! thanks for the contribution!!! 👏 🙏 

This PR needs a follow up which is change the "Blog" handle to the domain (which might need some massage if it has a long addressed as opposed to the domain only)

PR #172, opened 2021-05-11
Launch Readiness: Tailor Induction UI to Genesis Mode

brandonfancher:
We focused on providing a sensible experience and flow for inviters, invitees and witnesses. But after our small genesis induction user test, it was evident we didn't pay as much attention to invitees in Genesis mode. They were seeing UI as if the community were active

This addresses that. (Although we're in dire need of a good refactor of these flows; they've evolved organically and are difficult to reason about at the moment...but that will come later, together with some good cypress tests.)

This also addresses some of the other user feedback we received and witnessed during our test genesis induction yesterday.

### DIFF
--- a/packages/webapp/src/inductions/components/induction-step-endorsement.tsx
+++ b/packages/webapp/src/inductions/components/induction-step-endorsement.tsx
@@ -332,6 +332,7 @@ const DonationForm = ({
     setReviewStep,
     submitDonation,
 }: DonationFormProps) => {
+    const [isProfileReviewed, setReviewedProfile] = useState(false);
     return isInvitee ? (
         <div className="space-y-3">
             <Text>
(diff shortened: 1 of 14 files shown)


packages/webapp/src/inductions/components/induction-step-endorsement.tsx
```diff
@@ -297,46 +319,67 @@ const EndorsingForm = ({
...
             <Text>
-                This is your last chance to review your profile for completeness
-                and accuracy. If anything needs to be corrected,{" "}
+                This is your last chance to review your profile below for
+                completeness and accuracy. If anything needs to be corrected,{" "}
                 <Link onClick={() => setReviewStep("profile")}>click here</Link>
-                . Keep in mind that modifying your profile will require your
-                endorsers to submit their endorsements again.
+                .
+                {isCommunityActive &&
```
brandonfancher, line 343: When the time comes (soon) to refactor some of this...all these ternary checks and branching render logic indicates to me we should just make separate components for many of these views that have inviter/invitee/endorser/isCommunityActive roles...even if some of the text is shared among them.

packages/webapp/src/pages/induction/[id].tsx
```diff
@@ -67,6 +89,7 @@ export const InductionDetailsPage = () => {
                     <InductionStepEndorsement
                         induction={induction}
                         endorsements={endorsements}
+                        isCommunityActive={isCommunityActive}
```
Comment (brandonfancher) [inline packages/webapp/src/pages/induction/[id].tsx:92]: A good refactor will also get rid of a lot of this prop drilling.

packages/webapp/src/inductions/components/induction-step-endorsement.tsx
```diff
@@ -151,15 +152,29 @@ export const InductionStepEndorsement = (props: Props) => {
...
+    };
+
+    const getInductionJourneyStep = () => {
+        if (!props.isCommunityActive) {
+            return 2;
+        } else if (isFullyEndorsed) {
+            return 4;
+        }
+        return 3;
+    };
```
sparkplug0025, line 171: I think if we replace the step number with the actual step title we get a better readability. (Not for this PR though, should be a nice good first issue :) )

packages/webapp/src/inductions/components/induction-step-endorsement.tsx
sparkplug0025: yes! agreed.

sparkplug0025 (approved):
lgtm :shipit:

PR #174, opened 2021-05-12
Induction Journey Fixes

brandonfancher:
* "Ready to Join Eden" showed the induction journey graphic from the perspective of the Inviter instead of the Invitee. Fixed.
* Witnesses saw the induction journey graphic from the perspective of the Invitee instead of the Inviter on the "Waiting for Invitee" page. Fixed.
* Clarify invitation table header language.
* Added a TODO inline as a note for upcoming refactor.


brandonfancher:
These are all issues that occur after community is activated...not during Genesis, btw.

cool-ant (approved):

PR #194, opened 2021-05-17
IPFS API Endpoint

sparkplug0025:
Adds an endpoint that allows uploading files with a `inductprofil` or `inductvideo` action.


packages/webapp/.env
```diff
@@ -17,3 +17,6 @@ NEXT_PUBLIC_APP_NAME = "Eden Community App"
 NEXT_PUBLIC_EDEN_CONTRACT_ACCOUNT = "test.edev"
 NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT="10.00000000 WAX"
 # NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT="10.0000 EOS"
+IPFS_PINATA_JWT="<set-pinata-jwt-here>"
+IPFS_PINATA_API="https://api.pinata.cloud/psa"
+IPFS_UPLOAD_ENDPOINT_URL="https://api.pinata.cloud/pinning/pinFileToIPFS"
```
brandonfancher, line 22: Are we using this anywhere? Looks like we're just using the Pinning Services API instead.

packages/webapp/src/config.ts
```diff
@@ -72,3 +72,19 @@ export const chainConfig = {
...
+        !process.env.IPFS_PINATA_JWT ||
+        !process.env.IPFS_UPLOAD_ENDPOINT_URL)
+) {
+    throw new Error("Missing Config Secrets are not set");
+}
+
+export const ipfsConfig = {
+    pinataApi: process.env.IPFS_PINATA_API || "",
+    pinataJwt: process.env.IPFS_PINATA_JWT || "",
+    uploadEndpointUrl: process.env.IPFS_UPLOAD_ENDPOINT_URL || "",
```
brandonfancher, line 89: Same question here.

packages/webapp/.env
brandonfancher: Removed.

packages/webapp/src/config.ts
brandonfancher: Removed.

brandonfancher (approved):
Awesome!

PR #196, opened 2021-05-17
Adds an Upload File button for Image and adjusts the Preview

sparkplug0025:


brandonfancher:
Wow. @sparkplug0025 So am I reading this right? Infura allows for free IPFS uploads...so we client-side upload to Infura and then server-side pin to pinata.cloud (assuming the CID matches what's in the transaction)? And that helps us get around Vercel's 60-second serverless API function execution limit of 60 seconds?

brandonfancher (changes requested):
This is great! I set up a paid Pinata.cloud account for us (see your Signal) and set the env vars in Vercel. Tested locally and the preview deployment--working great. Just a few recommendations before we merge:

* I haven't rebased this yet...you'll of course want to do that. (And see my commit removing the unused env var.)
* Submitting the profile seemed to take around 15 seconds every test I did. This is totally fine, but as a user, it made me feel like the UI might have frozen up. Can we add a spinner icon to the submitting button...or some other moving indication that it's working/loading?
* Can we add the ? icon/button that redirects to Notion back in? Before we push this out to production, we'll just modify that page to remove the pinata.cloud instructions and add a few more guidelines (square, max file size, etc.) Hopefully soon, we can just make that a modal in the UI or something.

Once that's in place, I think we're good...even for pushing this out into production. It's a big improvement over the manual pinata approach. And I'm looking forward to seeing this working for video uploads too!

sparkplug0025:
> Wow. @sparkplug0025 So am I reading this right? Infura allows for free IPFS uploads...so we client-side upload to Infura and then server-side pin to pinata.cloud (assuming the CID matches what's in the transaction)? And that helps us get around Vercel's 60-second serverless API function execution limit of 60 seconds?

Actually 60 seconds is a good point, but not only that, Vercel has a hard payload limit at 5mb. That's why I needed to use something like this! But yeah, nice free service huh?

- Link is added back
- whoa... 15 seconds never happened to me, usually it's almost immediate (1-2 seconds) -- added the loading spinner in any case!

brandonfancher (approved):

PR #200, opened 2021-05-17
Post-footer home cleanup. Improve action button and CTA.

brandonfancher:
The Home page was looking a bit bare after moving links down into the footer. (heck, it was bare even before that!) This splits the information up a bit...makes it feel more full and easier on the eyes. I think?

Let me know what you think. (Compare eden-dev.vercel.app with this branch's deploy.)

I'm also improving the way `ActionButton` handles external links in this PR...similar to prior improvements to `Link`.

### DIFF
--- a/packages/webapp/src/_app/ui/action-button.tsx
+++ b/packages/webapp/src/_app/ui/action-button.tsx
@@ -63,12 +63,7 @@ export const ActionButton = ({
     const cursorClass = disabled ? "cursor-not-allowed" : "cursor-pointer";
     const buttonClass = `${baseClass} ${size} ${widthClass} ${colorClass} ${cursorClass} ${className}`;
 
-    let targetProps = {
-        target,
-        rel: isExternal ? "noopener noreferrer" : undefined,
-    };
-
-    if (onClick) {
+    if (isSubmit || onClick) {
         return (
             <button
                 onClick={onClick}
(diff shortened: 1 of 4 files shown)


packages/webapp/src/_app/ui/action-button.tsx
```diff
@@ -56,22 +60,37 @@ export const ActionButton = ({
...
-            className={buttonClass}
-            disabled={disabled}
-        >
-            {children}
-        </button>
+        <NextLink href={href}>
+            <a className={buttonClass} {...targetProps}>
+                {children}
+            </a>
+        </NextLink>
```
sparkplug0025, line 94: I think this change will break submit buttons (fill all the fields and press enter). Also I don't think it's a good idea to replace the button semantics with a link inside forms.

sparkplug0025: ohh nevermind... I just saw the button being returned above!

sparkplug0025 (approved):

packages/webapp/src/_app/ui/action-button.tsx
sparkplug0025: nevermind again lol... the button will be reeturned only if onclick is present, but submit buttons don't always have it: https://github.com/eoscommunity/Eden/blob/main/packages/webapp/src/inductions/components/induction-profile-form.tsx#L245

sparkplug0025 (dismissed):
See the comment about Submit buttons

packages/webapp/src/_app/ui/action-button.tsx
brandonfancher: yeah, good catch!

PR #201, opened 2021-05-18
Add polish

Velua:
- Typed the response we get from contract table requests
- Found an opportunity to fire two promises at the same time rather than wait for one to resolve before starting the other.
- Utilised account names as a key in one of the lists as we know for sure that those will be unique. 
- Adopted a IPFS url generator function, just a bit cleaner and will let us adopt another IPFS gateway which may have a different URL structure than the current one in place, did the same thing for the Block explorer account links.


packages/webapp/src/members/components/members-grid.tsx
```diff
@@ -21,8 +21,8 @@ export const MembersGrid = ({ members, dataTestId }: Props) => {
     return (
         <div className={containerClass} data-testid={dataTestId}>
             {(members.length &&
-                members.map((member, index) => (
-                    <MemberSquare key={index} member={member} />
+                members.map((member) => (
+                    <MemberSquare key={member.account} member={member} />
```
sparkplug0025, line 25: This one is kinda dangerous because we might reuse this MembersGrid to show cards (with different NFTs#) to show the sales/auctions for a single member... I'm fine with merging it now, but we just need to be careful to not reuse the grid in case same member cards is repeated.

packages/webapp/src/config.ts
```diff
@@ -73,6 +73,11 @@ export const chainConfig = {
     rpcEndpoints: [rpcEndpoint],
 };
 
+
+export const ipfsUrl = (ipfsHash: string) => `${ipfsBaseUrl}/${ipfsHash}`;
+export const explorerAccountUrl = (accountName: string) =>
+    `${blockExplorerAccountBaseUrl}/${accountName}`;
+
```
sparkplug0025, line 80: I'd like to keep this file clean with only the config settings. let's move this to _app/utils/config-helpers.ts or something. (I'm working on it!)

packages/webapp/src/_app/eos/api.ts
```diff
@@ -20,11 +26,11 @@ export const getRow = async <T>(
         : undefined;
 };
 
-export const getTableRows = async (
+export const getTableRows = async <T = any>(
```
sparkplug0025, line 29: love it! ❤️

PR #202, opened 2021-05-18
Add file input to upload videos to IPFS

sparkplug0025:
image

depends on #196

### DIFF
--- a/packages/webapp/src/inductions/components/induction-step-video.tsx
+++ b/packages/webapp/src/inductions/components/induction-step-video.tsx
@@ -16,7 +16,10 @@ import {
 } from "../utils";
 import { Induction } from "../interfaces";
 import { setInductionVideoTransaction } from "../transactions";
-import { InductionVideoForm } from "./induction-video-form";
+import {
+    InductionVideoForm,
+    VideoSubmissionPhase,
+} from "./induction-video-form";
 import {
     InductionJourneyContainer,
     InductionRole,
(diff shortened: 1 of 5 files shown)


brandonfancher (changes requested):
A couple things I noticed:

* video rerenders on window (re)focus (e.g., cmd tab)
* submit button appears as if it's enabled before a video is selected; should be disabled until video selection

brandonfancher:
I do have one usability concern...and I don't think it will be difficult to fix. This is even worse if the user is using Scatter, since the signing window often just appears _under_ the browser window. 

When the user submits, three things happen:
1. video uploads (could take a while)
2. user is prompted to sign (could prompt _under_ browser window and user may not notice)
3. we pin it...which can take another 5-10 seconds.

The user can miss step 2...and they might refresh thinking it's stuck or the transaction could expire when they finally realize it's time to sign.

If we can change the button text depending on the step their at:
1. Uploading video...
2. Waiting for you to sign...
3. Finishing up...

I think that minor change would go a long way to reassure the user everything's working and to prompt them to look for the signing window when it's time.

packages/webapp/src/inductions/components/induction-step-video.tsx
```diff
@@ -34,20 +39,31 @@ export const InductionStepVideo = ({
...
+                broadcast: false,
             });
             console.info("inductvideo trx", signedTrx);
+
+            setVideoSubmissionPhase("finishing");
+            await uploadIpfsFileWithTransaction(signedTrx, videoHash);
+
             setSubmittedVideo(true);
         } catch (error) {
             onError(error, "Unable to set the induction video");
```
brandonfancher, line 69: We need to `setVideoSubmissionPhase()` back to undefined here. Otherwise, if an error occurs or the user chooses not to sign, we don't get stuck like this:

image

brandonfancher (changes requested):

packages/webapp/src/inductions/components/induction-step-video.tsx
brandonfancher: I pushed the fix.

brandonfancher (approved):

brandonfancher (changes requested):
I just realized...neither of us read the words _around_ the video. They still say to upload the IPFS and paste CID. haha. I'll fix that.

brandonfancher (approved):

PR #208, opened 2021-05-18
Increase post-donate delay to give chain more time.

### LINKED ISSUE #207: 🎨 After donation, user sometimes sees non-member screen
Feedback from Robrigo: After `inductdonat`, users sometimes get the "This account is not an active Eden member" screen.

Introduce a delay so the app gives the chain enough time to catch up.

brandonfancher:
Fixes #207 (or attempts to).


brandonfancher:
If we want to get fancy, we can poll for the presence of the member record. But this should be good enough for now.

PR #213, opened 2021-05-19
add Dark mode

Velua:


brandonfancher (changes requested):
@Velua Sorry it's taken so long for us to get to reviewing this PR. Thanks for the contribution! The dark mode, as you've implemented it, is really easy on the eyes.

A few items:
* This will need a rebase.
* I like the dark/light mode toggle, but it needs to be centered vertically with the account text and logout button (it's slightly lower than those right now.) 
image
* What do you think about simply changing the text and stroke of the Eden logo from black in light mode to white in dark mode? This could be done directly to the SVG code, or you could pull in the white logo SVG here: https://www.notion.so/edenos/Eden-Logos-for-Download-dd13531e5bf94738a3b90f2efffd12b6
* The text in the Invitation Steps is blending in with the dark background. And I think the inside of the grey circles should match the background color (or simply be transparent?): 
image
* The NFT #X chip background is lost in dark mode: 
image

If you could comb through and look for anything else like that, that'd be great.

At a higher level, what do you think about setting this up to default to the user's OS settings at the time? Or maybe it defaults to OS settings with a manual override? We could ship without that, but might be nice to have.

brandonfancher:
Closing for now, as the feedback wasn't acted on and main has diverged so much. We will refer back to this when we work on dark mode again.

Issue #198, opened 2021-05-17
🎨 📟 Implement IPFS Pinning Limits


brandonfancher:
We're implementing IPFS uploading from our frontend (instead of telling users to go to Pinata.cloud and do it themselves.) This is how we do it:
1. User selects image or video and frontend uploads it to Infura's public IPFS gateway.
2. Transaction is sent to our backend serverless function, which inspects and does some integrity checks on the IPFS CID and then pins the file to our Pinata.cloud account.
Right now, there are no file size checks. A user could mistakenly upload a 100MB image, and we'd happily pin it in Pinata and pay for that storage.

We should implement sane limits by:
1. Check file size in the frontend and error out if image/video is too large. 
2. Because users could get around that if they want, inspect the file metadata in our serverless function prior to pinning it to Pinata.cloud and refuse to pin anything larger than our limits.

sparkplug0025:
Profile Images: 1mb
Videos: 100mb

PR #214, opened 2021-05-19
Add IPFS max size check

sparkplug0025:
Completes #198

### DIFF
--- a/packages/webapp/src/_api/handlers/ipfs-upload-handler.ts
+++ b/packages/webapp/src/_api/handlers/ipfs-upload-handler.ts
@@ -1,8 +1,27 @@
 import { eosDefaultApi, eosJsonRpc } from "_app";
-import { edenContractAccount, ipfsConfig } from "config";
+import { edenContractAccount, ipfsApiBaseUrl, ipfsConfig } from "config";
 
 import { BadRequestError, InternalServerError } from "../error-handlers";
 import { IpfsPostRequest } from "../schemas";
+
+interface ActionIpfsData {
+    cid: string;
+    contract: string;
+    action: string;
+}
+
+interface ValidUploadActions {
+    [contract: string]: {
+        [action: string]: { maxSize: number };
+    };
+}
+
+const VALID_UPLOAD_ACTIONS: ValidUploadActions = {
+    [edenContractAccount]: {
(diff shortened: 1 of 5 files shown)


packages/webapp/src/_api/handlers/ipfs-upload-handler.ts
```diff
@@ -1,23 +1,44 @@
...
+interface ValidUploadActions {
+    [contract: string]: {
+        [action: string]: { maxSize: number };
+    };
+}
+
+const VALID_UPLOAD_ACTIONS: ValidUploadActions = {
+    [edenContractAccount]: {
+        inductprofil: { maxSize: 1_000_000 },
+        inductvideo: { maxSize: 100_000_000 },
```
brandonfancher, line 22: Should we move these into the config?

packages/webapp/src/_app/hooks/forms.ts
```diff
@@ -42,5 +43,15 @@ export const handleFileChange = async (
         return onError(new Error(`You can only select ${mimePrefix} files`));
     }
 
+    if (file.size > maxSize) {
+        return onError(
+            new Error(
+                `The file cannot be larger than ${Math.floor(
+                    maxSize / 1_000_000
```
brandonfancher, line 50: Here too.

packages/webapp/src/inductions/components/induction-profile-form.tsx
```diff
@@ -97,7 +97,12 @@ export const InductionProfileForm = ({
                     accept="image/*"
                     label="select an image file"
                     onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
-                        handleFileChange(e, "image", setUploadedImage)
+                        handleFileChange(
+                            e,
+                            "image",
+                            1_000_000,
```
brandonfancher, line 103: Here too.

packages/webapp/src/_api/handlers/ipfs-upload-handler.ts
brandonfancher: ...and reference them in the three places they're also used below?

sparkplug0025: good idea! done!

brandonfancher (approved):

Issue #173, opened 2021-05-11
🎨 Minor refactor task in InductionJourneyContainer


brandonfancher:
I don't love this `vAlign` thing. Quick fix for now, but worth a refactor.

_Originally posted by @brandonfancher in https://github.com/eoscommunity/Eden/pull/172#r630567458_

PR #218, opened 2021-05-19
Induction Flow Reflow/Refactor

brandonfancher:
The induction flow has become some gnarly spaghetti! 🍝 Extremely brittle. I'm afraid to touch it. So I'm going just going to dive in.

In this refactor, I switch first on the role of the user (inviter, witness, invitee, other, unauthenticated, etc.) and _then_ on the induction status. It's much easier to think in terms of the user, IMHO. Want to change invitee's instructions on the donate status? `Invitee` role > `Donate` status. Easier to reason about.

### What's in this?
* UI code reflow/refactor for entire invitation/induction flow.
* Fixes bugs, jarring rerenders on window refocus, and state transition flakiness.


brandonfancher:
@sparkplug0025 I'm going ahead and marking this as ready for your review. I'm still doing some manual testing on my end, but I don't anticipate any big changes from what's here now.

I suggest starting by pulling this down and familiarizing yourself with the new file/dir structure for inductions. Then dig into the code starting at the [id].tsx inductions page.

brandonfancher:
Fixes #173

packages/webapp/src/_app/ui/action-button.tsx
```diff
@@ -24,12 +24,12 @@ export enum ActionButtonSize {
...
-    NEUTRAL = "bg-gray-50 text-gray-800 hover:bg-gray-200",
-    INDUCTION_STATUS_PROFILE = "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
-    INDUCTION_STATUS_CEREMONY = "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
-    INDUCTION_STATUS_ACTION = "bg-green-500 text-white hover:bg-green-600",
+    Primary = "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
+    Disabled = "border-gray-400 bg-gray-300 text-gray-500",
+    Neutral = "bg-gray-50 text-gray-800 hover:bg-gray-200",
+    InductionStatusProfile = "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
+    InductionStatusCeremony = "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
+    InductionStatusAction = "bg-green-500 text-white hover:bg-green-600",
```
sparkplug0025, line 32: Since we are on it... It's a good time to consider getting rid of Enums for React props. The reason for that is that anytime we want to use the `ActionButton` we also need to import `ActionButtonSize` and `ActionButtonType` which kinda sucks... A solution for that with union types is: 

```tsx
// inside action-button.tsx 
type ActionButtonType =
    | "primary"
    | "disabled"
    | "neutral"
    | "InductionStatusProfile"
    | "InductionStatusCeremony"
    | "InductionStatusAction";
const ActionButtonTypes: { [key in ActionButtonType]: string } = {
    primary: "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
    disabled: "border-gray-400 bg-gray-300 text-gray-500",
    neutral: "bg-gray-50 text-gray-800 hover:bg-gray-200",
    InductionStatusProfile:
        "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
    InductionStatusCeremony:
        "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
    InductionStatusAction: "bg-green-500 text-white hover:bg-green-600",
};

// using action button externally:
<ActionButton type="primary" /> // reacty!
```

packages/webapp/src/inductions/api/eden-contract.ts
```diff
@@ -15,15 +15,16 @@ const INDEX_BY_INVITER = 3;
 const INDEX_BY_ENDORSER = 2;
 const INDEX_BY_INDUCTION = 3;
 
-export const getInductionWithEndorsements = async (
-    inductionId: string
-): Promise<
+export const getInductionWithEndorsements = async ({
+    queryKey,
+}: any): Promise<
```
sparkplug0025, line 20: this one abstracts away how to use this function... not good...
before I knew that if  I want to "get induction with endorsements" I could simply pass the induction Id. now I have no idea what I should pass to this function... 
Also, I can't understand what is the purpose of this _key since we are not using it anywhere inside this function. I think this abstraction should live outside... eg:

```
export const useGetInductionWithEndorsements = (inductionId: string) =>
    useQuery(["induction", inductionId], () => getInductionWithEndorsements(inductionId), {
        // gets rid of bugs cause by refetching right after anchor/scatter signing
        refetchOnWindowFocus: false,
    });
```

With that we don't need to touch the getInductionWithEndorsements. And getInductionWithEndorsements can also be used outside react-query if we want to call it from our backend.

packages/webapp/src/inductions/utils.ts
```diff
@@ -23,20 +27,28 @@ export const convertPendingProfileToMemberData = (
...
         .isBefore(dayjs());
 
-    return isExpired
-        ? InductionStatus.expired
-        : !induction.new_member_profile.name
-        ? InductionStatus.waitingForProfile
+    if (isExpired) return InductionStatus.Expired;
+
+    const isWaitingForDonation =
+        endorsements?.every((e) => e.endorsed === 1) ?? false;
```
sparkplug0025, line 43: Could be simplified by: `Boolean(endorsements?.every((e) => e.endorsed === 1))`

packages/webapp/src/inductions/hooks/index.ts
```diff
@@ -0,0 +1,18 @@
...
+): InductionRole => {
+    const [ualAccount] = useUALAccount();
+    const userRole: InductionRole = useMemo(
+        () => getInductionUserRole(endorsements, ualAccount, induction),
+        [ualAccount, induction, endorsements]
+    );
+    return userRole;
+};
+
+export * from "./queries";
```
sparkplug0025, line 18: I like this query + hook refactor. But I would not mix logic + exports on the index file. All of our indexes files are exports. If you think they are small splitting them away would be silly we should remove the hooks folder and put them together in a hooks.ts file, until it grows and we can then create a folder (pretty much like you did with the `induction-journeys` which I love it :) )

packages/webapp/src/inductions/interfaces.ts
```diff
@@ -16,12 +16,23 @@ export interface Induction {
...
 }
 
+export enum InductionRole {
+    Inviter = "inviter",
+    Endorser = "endorser",
+    Invitee = "invitee",
+    Member = "member", // use?
+    NonMember = "non-member", // use?
+    Unauthenticated = "unauthenticated",
+    Unknown = "unknown",
```
sparkplug0025, line 26: out of curiosity, any reason why we need named enums? if it's not being used we could leave it as a plain enum....

packages/webapp/src/inductions/components/induction-journeys/common/induction-journey-container.tsx
```diff
@@ -1,123 +1,147 @@
...
 
-export enum InductionRole {
-    INVITEE = "invitee",
-    INVITER = "inviter",
-    GENESIS = "genesis",
-}
-
 interface Props {
-    role: InductionRole;
-    step: 1 | 2 | 3 | 4 | 5;
```
sparkplug0025, line 87: I love how you are using enums for the steps!

packages/webapp/src/inductions/components/induction-journeys/common/induction-journey-container.tsx
```diff
@@ -1,123 +1,147 @@
...
-            steps = GENESIS_INDUCTION_STEPS;
-            break;
-        case InductionRole.INVITEE:
-            steps = INVITEE_INDUCTION_STEPS;
-            break;
-        case InductionRole.INVITER:
-            steps = INVITER_INDUCTION_STEPS;
-            break;
+export const InductionJourneyContainer = ({ step, children }: Props) => {
+    function isStepIn<T>(steps: T) {
```
sparkplug0025, line 127: 🏹 arrow function? :)

packages/webapp/src/inductions/components/induction-journeys/common/waiting-for-video.tsx
```diff
@@ -0,0 +1,34 @@
...
+                <Text className="leading-normal">
+                    <span className="font-medium">
+                        Remember, this invitation is still open and expires in{" "}
+                        {getInductionRemainingTimeDays(induction)}.
+                    </span>{" "}
+                    If time runs out, you can always request another invitation.
+                </Text>
+                <Text className="leading-normal">
+                    In the meantime, review your Eden profile below.
+                </Text>
```
sparkplug0025, line 30: this text is written targeting the invitee, but this component is also being used on the third-party-journey. I think it's not a biggie...

packages/webapp/src/pages/induction/[id].tsx
```diff
@@ -1,128 +1,106 @@
...
+    }, [induction, endorsements, status, userRole]);
+
+    // TODO: We need a common loading spinner component.
+    if (isLoading) {
+        return (
+            <SingleColLayout title="Loading">
+                <Card title="Loading...">...</Card>
+            </SingleColLayout>
+        );
+    }
```
Comment (sparkplug0025) [inline packages/webapp/src/pages/induction/[id].tsx:81]: yeah, it's a good idea to get a common loading spinner... aside from that I see that here we are replacing the whole layout. can't we just have the same wrapper with the `Induction #` title?

packages/webapp/src/pages/induction/[id].tsx
```diff
@@ -1,128 +1,106 @@
...
-    , [ualAccount, endorsements]);
+    const induction = data?.induction;
+    const endorsements = data?.endorsements ?? [];
 
-    const status = getInductionStatus(induction);
+    const userRole = useInductionUserRole(endorsements, induction);
+    const status = getInductionStatus(induction, endorsements);
 
-    const renderInductionStep = useMemo(() => {
+    const renderInductionJourney = useMemo(() => {
```
Comment (sparkplug0025) [inline packages/webapp/src/pages/induction/[id].tsx:41]: I think it's a good opportunity to extract this whole memo to another component container... this page component would be way cleaner

packages/webapp/src/pages/induction/[id].tsx
```diff
@@ -1,128 +1,106 @@
...
 
-    // TODO: Consider deriving the user's role here and return an enum of roles: INVITER, ENDORSER, INVITEE, MEMBER, EOS_USER, UNAUTHENTICATED.
-    // Almost every child component of this page cares about the role of the user in relation to the invite/induction. We can pass role down.
-    const isEndorser = useMemo(() => 
-         endorsements.some(
-            (endorsement) => endorsement.endorser === ualAccount?.accountName
-        )
-    , [ualAccount, endorsements]);
+    const induction = data?.induction;
+    const endorsements = data?.endorsements ?? [];
```
Comment (sparkplug0025) [inline packages/webapp/src/pages/induction/[id].tsx:36]: What would be the reason for `??` vs `||` on this particular case?

packages/webapp/src/pages/induction/[id].tsx
```diff
@@ -1,128 +1,106 @@
...
+                        endorsements={endorsements}
                         induction={induction}
-                        isEndorser={isEndorser}
+                        inductionStatus={status}
                     />
                 );
-            case InductionStatus.waitingForEndorsement:
+            case InductionRole.Member:
+            case InductionRole.Unknown:
+            case InductionRole.Unauthenticated:
```
Comment (sparkplug0025) [inline packages/webapp/src/pages/induction/[id].tsx:63]: replace with default... otherwise we need to maintain two places whenever we add a new Role to make sure it can view the ThirdPartyJourney

packages/webapp/src/_app/ui/action-button.tsx
brandonfancher: I like it. Let's do it, but we'll have a separate PR soon for cleaning up our shared components (Text, buttons, etc.) I only changed this here since I was already working with enums and wanted consistency with the capitalization.

packages/webapp/src/inductions/api/eden-contract.ts
brandonfancher: This is more of the `react-query` way of doing this: https://react-query.tanstack.com/guides/query-functions#query-function-variables. But I don't feel strongly about it either way.

If we plan to use `getInductionWithEndorsements` outside of `react-query`, then switching it back is prob not a bad idea.

packages/webapp/src/inductions/utils.ts
brandonfancher: Yeah, either works. I'll wrap with `Boolean`. 👍

packages/webapp/src/inductions/interfaces.ts
brandonfancher: I'll double check where/how we're using those enums. But plain enum is prob fine.

packages/webapp/src/inductions/components/induction-journeys/common/induction-journey-container.tsx
brandonfancher: Sure! I'm still fuzzy on it... is there a benefit one way or the other in cases like this?

packages/webapp/src/inductions/components/induction-journeys/common/waiting-for-video.tsx
brandonfancher: Yeah. In this case, I'm consciously having third parties read from the invitee's perspective. I think it works.

brandonfancher: Third parties will often become invitees themselves at some point...so this primes their expectations.

packages/webapp/src/pages/induction/[id].tsx
Comment (brandonfancher) [reply packages/webapp/src/pages/induction/[id].tsx:81]: Yeah, I don't see why not. I'll modify. 👍

packages/webapp/src/inductions/components/induction-journeys/invitee-journey.tsx
```diff
@@ -0,0 +1,156 @@
...
+                        setIsReviewingProfile={setIsReviewingProfile}
+                    />
+                </Container>
+            );
+        default:
+            return <p>Unknown error</p>;
+    }
+};
+
+export default InviteeJourney;
```
sparkplug0025, line 156: We could get it even simpler and readable with extracting the branching to it's own components, something like this (need to replace the any's though): 

```tsx
// ...

export const InviteeJourney = ({
    endorsements,
    induction,
    inductionStatus,
}: Props) => {
    const [submittedProfile, setSubmittedProfile] = useState(false);
    const [isReviewingProfile, setIsReviewingProfile] = useState(false);

    if (submittedProfile) {
        return <SubmittedProfileStep />;
    }

    if (isReviewingProfile) {
        return (
            <ProfileStep
                induction={induction}
                isReviewingProfile={isReviewingProfile}
                setSubmittedProfile={setSubmittedProfile}
            />
        );
    }

    switch (inductionStatus) {
        case InductionStatus.PendingProfile:
            return (
                <ProfileStep
                    induction={induction}
                    isReviewingProfile={isReviewingProfile}
                    setSubmittedProfile={setSubmittedProfile}
                />
            );
        case InductionStatus.PendingCeremonyVideo: // not possible in Genesis mode
            return <PendingCeremonyVideoStep induction={induction} />;
        case InductionStatus.PendingEndorsement: // not possible in Genesis mode
            return (
                <PendingEndorsementStep
                    induction={induction}
                    endorsements={endorsements}
                    setIsReviewingProfile={setIsReviewingProfile}
                />
            );
        case InductionStatus.PendingDonation:
            return (
                <PendingDonateStep
                    induction={induction}
                    endorsements={endorsements}
                    setIsReviewingProfile={setIsReviewingProfile}
                />
            );
        default:
            return <p>Unknown error</p>;
    }
};

export default InviteeJourney;

const SubmittedProfileStep = () => {
    const { data: isCommunityActive } = useIsCommunityActive();
    return (
        <Container
            step={
                isCommunityActive
                    ? InductionStepInvitee.PendingVideoAndEndorsements
                    : InductionStepGenesis.Donate
            }
        >
            <InductionProfileSubmitConfirmation
                isCommunityActive={isCommunityActive}
            />
        </Container>
    );
};

const ProfileStep = ({
    induction,
    isReviewingProfile,
    setSubmittedProfile,
}: any) => {
    const { data: isCommunityActive } = useIsCommunityActive();
    return (
        <Container
            step={
                isCommunityActive
                    ? InductionStepInvitee.Profile
                    : InductionStepGenesis.Profile
            }
        >
            <InductionProfileFormContainer
                induction={induction}
                isReviewingProfile={isReviewingProfile}
                setSubmittedProfile={setSubmittedProfile}
            />
        </Container>
    );
};

const PendingCeremonyVideoStep = ({ memberData, induction }: any) => (
    <Container
        step={InductionStepInvitee.PendingVideoAndEndorsements}
        memberPreview={memberData}
    >
        <WaitingForVideo induction={induction} />
    </Container>
);

const PendingEndorsementStep = ({
    induction,
    endorsements,
    setIsReviewingProfile,
}: any) => {
    const memberData = convertPendingProfileToMemberData(induction);

    return (
        <Container
            step={InductionStepInvitee.PendingVideoAndEndorsements}
            memberPreview={memberData}
        >
            <Heading size={1} className="mb-2">
                Endorsements
            </Heading>
            <InductionExpiresIn induction={induction} />
            <EndorsementsStatus endorsements={endorsements} />
            <div className="space-y-3">
                <Text>To continue, all witnesses must endorse.</Text>
                <Text>
                    Now is a good time to review your profile information below.
                    If anything needs to be corrected,{" "}
                    <Link onClick={() => setIsReviewingProfile(true)}>
                        click here to make those adjustments.
                    </Link>{" "}
                    Keep in mind that any modifications to your profile will
                    reset any endorsements.
                </Text>
            </div>
        </Container>
    );
};

const PendingDonateStep = ({
    induction,
    endorsements,
    setIsReviewingProfile,
}: any) => {
    const memberData = convertPendingProfileToMemberData(induction);
    const { data: isCommunityActive } = useIsCommunityActive();

    return (
        <Container
            step={
                isCommunityActive
                    ? InductionStepInvitee.Donate
                    : InductionStepGenesis.Donate
            }
            memberPreview={memberData}
        >
            <Heading size={1} className="mb-2">
                Pending donation
            </Heading>
            <InductionExpiresIn induction={induction} />
            {isCommunityActive && (
                <EndorsementsStatus endorsements={endorsements} />
            )}
            <InductionDonateForm
                induction={induction}
                isCommunityActive={isCommunityActive}
                setIsReviewingProfile={setIsReviewingProfile}
            />
        </Container>
    );
};
```

packages/webapp/src/pages/induction/[id].tsx
Comment (brandonfancher) [reply packages/webapp/src/pages/induction/[id].tsx:41]: Good call.

packages/webapp/src/inductions/components/induction-journeys/inviter-witnesses/video-form.tsx
```diff
@@ -39,13 +39,13 @@ export const InductionVideoForm = ({
         setIsLoading(false);
     };
 
-    const getVideoUrl = () => {
+    const videoUrl = useMemo(() => {
```
sparkplug0025, line 42: 👏

packages/webapp/src/pages/induction/[id].tsx
Comment (brandonfancher) [reply packages/webapp/src/pages/induction/[id].tsx:36]: Nullish coalescing operator `??` here is more idiomatic. I only include `?? []` because I want it to fall back to the empty array _specifically_ if `data` is `undefined`.  Until `??` was introduced, we only had `||` for this, but the latter is less precise: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

packages/webapp/src/inductions/components/induction-journeys/invitee-journey.tsx
brandonfancher: Oooh, I _like_ it!

packages/webapp/src/inductions/components/induction-journeys/inviter-witnesses/video-form.tsx
brandonfancher: Yeah. it was still re-rendering when you hit the submit button. This fixed that.

packages/webapp/src/inductions/components/induction-journeys/inviter-witnesses/invite-form.tsx
```diff
@@ -1,5 +1,4 @@
 import { FormEvent, useState } from "react";
-
```
sparkplug0025, line 2: so, my rationale in imports is split the third party packages from our own packages... I always try to split in three, or even four, blocks:
```tsx
// import from npm packages 1
// import from npm packages 2
// import from npm packages 3

// import from our global packages, eg: "_app" 1
// import from our global packages, eg: "_app" 2
// import from our global packages, eg: "_app" 3

// import from siblings & relational files: eg "./ or "../xpto" 1
// import from siblings & relational files: eg "./ or "../xpto" 2
// import from siblings & relational files: eg "./ or "../xpto" 3

// import our own static assets, eg: svg, css, img etc. 1
// import our own static assets, eg: svg, css, img etc. 2
// import our own static assets, eg: svg, css, img etc. 3
```

If you don't like it, and prefer to keep them mixed, it's fine for me. I usually find this separation useful for reading, easy to quick scan and understand dependencies.

brandonfancher: I like your rationale. I'll comb through the files in this PR and adjust to follow it. 👍

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
```diff
@@ -0,0 +1,193 @@
...
+                <Container
+                    step={InductionStepInviter.VideoAndEndorse}
+                    memberPreview={memberData}
+                >
+                    <Heading size={1} className="mb-2">
+                        Endorsements
+                    </Heading>
+                    <InductionExpiresIn induction={induction} />
+                    <EndorsementsStatus endorsements={endorsements} />
+                    {userEndorsementIsPending ? (
```
sparkplug0025, line 134: I think that we should display the recommend review even if the endorser has not endorsed yet. Actually I think it's the correct behavior, in case the endorser has found an issue in the video and want's to fix it before endorsing it. He would need to endorse it and quickly click in review video (it could be worse if he is the last endorser).

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
```diff
@@ -0,0 +1,193 @@
...
+const Container = ({ step, memberPreview, children }: ContainerProps) => (
+    <>
+        <InductionJourneyContainer step={step}>
+            {children}
+        </InductionJourneyContainer>
+        {memberPreview && <MemberCardPreview memberData={memberPreview} />}
+    </>
+);
+
+const RecommendReview = ({
```
sparkplug0025, line 35: can we push it to the end of the file or extract it to its own component? it seems unnatural to open the file and do not see the main File component in it (`InviterWitnessJourney`).

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
```diff
@@ -0,0 +1,193 @@
...
+                        </Text>
+                    )}
+                </Container>
+            );
+        default:
+            return <p>Unknown error</p>;
+    }
+};
+
+export default InviterWitnessJourney;
```
sparkplug0025, line 193: I think like I recommended above for the invitee journey this could benefit from extracting the branching components and logic to its own components.

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
brandonfancher: A recommendation and link to review the video is already inside the `<InductionEndorsementForm />` component here. So it appears in either case already.

I will check, however, to see if that component can also just use `<RecommendReview />`. There might be some differences in wording that makes the one inside `<InductionEndorsementForm />` special.

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
brandonfancher: Yeah, I can push it to the end of the file. If more than just this file is referencing it, I'll break it out into its own file.

packages/webapp/src/inductions/components/induction-journeys/third-party-journey.tsx
```diff
@@ -0,0 +1,120 @@
...
+                        </Text>
+                    )}
+                </Container>
+            );
+        default:
+            return <p>Unknown error</p>;
+    }
+};
+
+export default ThirdPartyJourney;
```
sparkplug0025, line 120: same comment about extracting the branches to its own components like suggested above for invitee and inviter journeys. although here the branches logics is very thin, not that much benefit by doing that, but still great

packages/webapp/src/inductions/components/induction-journeys/common/induction-journey-container.tsx
sparkplug0025: not really, just keeping the codebase consistent

packages/webapp/src/inductions/api/eden-contract.ts
sparkplug0025: Yeah I get it is something for react-query. The problem is when we mix a simple data fetch with something that does not belong to it, meaning, from outside. This would be the first function where we are leaking something from our "Data Management SDK" to one of our API calls. And this is not desirable at all, because if we try to use it in our backend, or if we want to get rid of React-Query or maybe reuse it with another lib (wrapping with graphql or whatsoever), we just can't. 

See our [Members Abstractions](https://github.com/eoscommunity/Eden/blob/main/packages/webapp/src/members/api/members.ts) or even the other api calls to our EOS tables in this very same file... They are pure abstractions of how we call external apis. It has zero knowledge about our fetch/state mechanism. Can be reused anywhere, with any library.

packages/webapp/src/inductions/interfaces.ts
sparkplug0025: Yeah I noticed that we needed named enums in the genesis steps comparisons since we can't have `instanceof` enums. But here I don't think so... so cleaning it up is just easier to maintain

packages/webapp/src/pages/induction/[id].tsx
Comment (sparkplug0025) [reply packages/webapp/src/pages/induction/[id].tsx:36]: Got it... Let's say something wonky happens in the API and it returns `""`, or `0`, or `false`. I know it's completely wrong and we can't blame the code, but `||` would fallback to an empty array since it has a broader range of "empty" values vs `??` that only considers `null` or `undefined`. In that case `??` would break our code whereas `||` would keep it running. Even though, again, not our code fault, it's just that I've seen enough wonkiness with various APIs that I prefer to err on the safe side. I'm fine with either way you want to go with it, just some food for thoughts :)

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
sparkplug0025: Awwww okay, got it! you are absolutely right, I just re-checked it, and I missed it!

sparkplug0025 (approved):
Leaving it approved for mergeability after the implemented suggestions are done!

packages/webapp/src/pages/induction/[id].tsx
Comment (brandonfancher) [reply packages/webapp/src/pages/induction/[id].tsx:36]: You make a good case. I'll switch to `||`.

packages/webapp/src/inductions/components/induction-journeys/inviter-witness-journey.tsx
brandonfancher: Yeah, I went ahead and consolidated those.

brandonfancher:
@sparkplug0025 Feedback implemented. I'm running out of time today; I'll test everything again tomorrow and merge in the morn.

PR #219, opened 2021-05-20
Add Members Pagination

sparkplug0025:
Solves #4

### DIFF
--- a/packages/webapp/src/_app/ui/action-button.tsx
+++ b/packages/webapp/src/_app/ui/action-button.tsx
@@ -26,7 +26,7 @@ export enum ActionButtonSize {
 export enum ActionButtonType {
     PRIMARY = "bg-blue-500 border-blue-500 text-white hover:bg-blue-600",
     DISABLED = "border-gray-400 bg-gray-300 text-gray-500",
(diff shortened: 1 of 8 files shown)


brandonfancher (approved):
This works well!

A couple of items for a follow-up PR:
* We should also paginate new members. I imagine we're going to have a big influx of new members after the Genesis community opens to the public. And having 100 entries at the top of the members list page will be weird if the group below it is paginated.
* At the largest screen size, two cards are dangling on a third row by themselves. This is okay for now...we really need to rearrange the cards anyway...we can prob solve that differently by tweaking card design.

I'm going to go ahead and ship this because it is a good improvement.

brandonfancher:
> This works well!
> 
> A couple of items for a follow-up PR:
> 
> * We should also paginate new members. I imagine we're going to have a big influx of new members after the Genesis community opens to the public. And having 100 entries at the top of the members list page will be weird if the group below it is paginated.
> * At the largest screen size, two cards are dangling on a third row by themselves. This is okay for now...we really need to rearrange the cards anyway...we can prob solve that differently by tweaking card design.
> 
> I'm going to go ahead and ship this because it is a good improvement.

Ah, I see that we already have an issue for New Member pagination: #3. Good!

PR #224, opened 2021-05-21
add new members pagination

sparkplug0025:
and also fixes the dangling cards! new members: 8 / members: 16 :)


brandonfancher (approved):
This is definitely an improvement. ✅

Hopefully soon, we'll redesign these member pages and cards, introduce search, etc.

Issue #143, opened 2021-05-06
Add Environment Variable for enabled Wallets


sparkplug0025:
UalProvider breaks localhost if Scatter is not detected

One of the new `UALProvider` authenticators is erroring out. In dev mode, it's loud (see below). I don't think it impacts user experience in prod mode though. But it is annoying in dev mode.

One of the new  authenticators is erroring out. In dev mode, it's loud (see below). I don't think it impacts user experience in prod mode though. But it is annoying in dev mode.

![image](https://user-images.githubusercontent.com/7911424/117218774-cfbd6c80-add1-11eb-94c5-dcc91eaf0393.png)
![image](https://user-images.githubusercontent.com/7911424/117218789-d6e47a80-add1-11eb-804b-6323be6be56e.png)

I think it's Scatter. Not sure. And possibly only happens when Scatter isn't installed on the machine.

_Originally posted by @brandonfancher in https://github.com/eoscommunity/Eden/issues/80#issuecomment-833097463_

Just confirmed, it happens when Scatter is not open in the local machine.

So, if we add an Environment variable for a list of enabled wallets we can override production/dev/local...
eg: 
- in prod & stage: NEXT_PUBLIC_ENABLED_WALLETS="ANCHOR,SCATTER,LEDGER"
- local: NEXT_PUBLIC_ENABLED_WALLETS="ANCHOR,LEDGER"

PR #234, opened 2021-05-24
Add Dynamic Wallets Config Vars

sparkplug0025:
Solves #143


packages/webapp/src/_app/eos/ual/EdenUALProvider.tsx
```diff
@@ -1,8 +1,19 @@
...
+const authenticators: any[] = [];
+if (availableWallets.indexOf("ANCHOR") >= 0) {
+    authenticators.push(anchor);
+}
+if (availableWallets.indexOf("SCATTER") >= 0) {
+    authenticators.push(scatter);
+}
+if (availableWallets.indexOf("LEDGER") >= 0) {
+    authenticators.push(ledger);
+}
```
brandonfancher, line 15: `availableWallets.includes("ANCHOR")` is more readable here, IMHO.

brandonfancher (approved):
A minor suggestion, but good either way! I like how we're leveraging `.env.local` to get around the local Scatter issue. Nice!

brandonfancher:
@sparkplug0025 Do you know if we need to add this new env var to Vercel, or will it just read it from the `.env` file?

packages/webapp/src/_app/eos/ual/EdenUALProvider.tsx
sparkplug0025: oooo `.includes`! nice!

sparkplug0025:
> @sparkplug0025 Do you know if we need to add this new env var to Vercel, or will it just read it from the `.env` file?

I added already yesterday!

PR #235, opened 2021-05-25
wasm debugging

tbfleming:
Builds on #212


tbfleming:
I guess I could resort to a macro which is used for each specialization to avoid partial specialization

swatanabe:
> I guess I could resort to a macro which is used for each specialization to avoid partial specialization

I think overall it's a win, since you only need two specializations (one for contracts, one for tests, right? or are there more that I missed).  You only need to specialize `parse_section`; `parse_module` and the rest are only there to call the modified `parse_section`.

swatanabe:
I think in the long run, changing eos-vm is the best option, of course.

programs/cltester/dwarf.cpp
```diff
@@ -0,0 +1,1812 @@
...
+      uint32_t version = 1;
+      jit_actions action_flag = jit_noaction;
+      jit_code_entry* relevant_entry = nullptr;
+      jit_code_entry* first_entry = nullptr;
+   };
+}  // namespace dwarf
+
+extern "C"
+{
+   void __attribute__((noinline, optnone)) __jit_debug_register_code(){};
```
swatanabe, line 1611: gcc 10 is optimizing out the call.

tbfleming: dang it

swatanabe: The gcc documentation of noinline suggests `asm("")`

tbfleming: There were 2 issues with gcc:
* 10 optimized out the call
* 9 and 10 produced `*** longjmp causes uninitialized stack frame ***: terminated`

I added workarounds for both.

programs/cltester/debug-eos-vm.hpp
```diff
@@ -0,0 +1,192 @@
...
+            }
+         }
+
+         offset_to_addr = instr_locs.data();
+         offset_to_addr_len = instr_locs.size();
+      }
+
+      void relocate(const void* new_base) { code_begin = new_base; }
+
+      // Cannot use most of the standard library as the STL is not async-signal-safe
```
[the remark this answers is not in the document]
tbfleming: removed

programs/cltester/dwarf.cpp
```diff
@@ -523,6 +592,146 @@ namespace dwarf
...
+   }
+
+   template <typename S>
+   void write_line_program(const info& info,
+                           const std::vector<jit_fn_loc>& fn_locs,
+                           const std::vector<jit_instr_loc>& instr_locs,
+                           const void* code_start,
+                           S& s)
+   {
+      std::optional<uint64_t> address = 0;
```
swatanabe, line 637: Why is `address` an `optional`?  I don't see any way that it can be set to `nullopt`.

programs/cltester/dwarf.cpp
```diff
@@ -928,8 +1176,221 @@ namespace dwarf
...
+      }
+   };
+
+   struct die_pattern
+   {
+      uint32_t tag = 0;
+      bool has_children = false;
+      std::vector<attr_form_value> attrs;
+
+      auto key() const { return std::tuple{tag, has_children, attrs}; }
```
swatanabe, line 1221: I would use tie to avoid copying the vector.

programs/cltester/dwarf.cpp
```diff
@@ -928,8 +1176,221 @@ namespace dwarf
...
+      auto result = strings.size();
+      strings.insert(strings.end(), s, s + strlen(s) + 1);
+      return result;
    }
 
+   template <typename T>
+   struct fill_later
+   {
+      size_t* offset = nullptr;
+   };
```
swatanabe, line 1197: I don't see anything that creates an attribute with this kind of value.

programs/cltester/dwarf.cpp
```diff
@@ -928,8 +1176,221 @@ namespace dwarf
...
+         if (show_generated_dies)
+            fprintf(stderr, "    DIE 0x%lx (%ld) subprogram %08x-%08x %016lx-%016lx %s\n",
+                    uint64_t(info_data.size()), uint64_t(i),
+                    info.wasm_code_offset + sub.begin_address,
+                    info.wasm_code_offset + sub.end_address, fn_begin + print_addr_adj,
+                    fn_end + print_addr_adj, sub.demangled_name.c_str());
+
+         die.tag = dw_tag_subprogram;
+         die.has_children = false;
+         if (sub.parent)
```
swatanabe, line 1349: Didn't we just check `sub.parent` and bail out above?

programs/cltester/main.cpp
```diff
@@ -1555,21 +1670,53 @@ static void run(const char* wasm, const std::vector<std::string>& args)
...
+      -h, --help
+
+            Show this message
+
+      -v, --verbose
+
+            Show detailed logging
+
+      -s contract.wasm debug.wasm
+      --subst contract.wasm debug.wasm
```
swatanabe, line 1685: I wonder if external debug info works: https://yurydelendik.github.io/webassembly-dwarf/#external-DWARF.  Also, is it possible to use RelWithDebugInfo?

programs/cltester/main.cpp
```diff
@@ -369,7 +407,61 @@ struct test_chain
...
+      if (auto it = state.cached_modules.find(code_hash); it != state.cached_modules.end())
+         return it->second;
+
+      if (auto it = state.codes.find(code_hash); it != state.codes.end())
+      {
+         IR::Module module;
+         auto dwarf_info =
+             dwarf::get_info_from_wasm({(const char*)it->second.data(), it->second.size()});
+         auto size = dwarf::wasm_exclude_custom({(const char*)it->second.data(), it->second.size()})
+                         .remaining();
```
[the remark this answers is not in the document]
tbfleming: I'll add a check to `wasm_exclude_custom`; clang puts them at the end

programs/cltester/dwarf.cpp
tbfleming: At one point I had the code partially completed to generate a tree for inlined code. I stripped most of it out to finish later if needed (clang still inlines in spots even with `-O1`; `-fno-inline-functions` doesn't even prevent it). I'll remove this.

programs/cltester/main.cpp
tbfleming: Clang in wasm mode doesn't produce dwarf at `-O3` or `-Os`. I suspect it's because it invokes `wasm-opt`

programs/cltester/main.cpp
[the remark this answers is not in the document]
tbfleming: changed

programs/cltester/main.cpp
```diff
@@ -369,7 +407,61 @@ struct test_chain
...
+         Serialization::MemoryInputStream stream(it->second.data(), size);
+         WASM::scoped_skip_checks no_check;
+         WASM::serialize(stream, module);
```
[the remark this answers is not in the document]
tbfleming: `module` and `stream` no longer used; removed

programs/cltester/dwarf.cpp
tbfleming: changed

programs/cltester/dwarf.cpp
tbfleming: changed

programs/cltester/dwarf.cpp
tbfleming: The code that used it is no longer present. removed.

programs/cltester/dwarf.cpp
tbfleming: removed

swatanabe (approved):

Issue #251, opened 2021-05-26
🎨 🛠 FE Component Refactors


brandonfancher:
* Replace instances of `Button` with `ActionButton` and rename `ActionButton` to `Button`
* Make components relying too heavily on TS enums as props use normal union types instead...so they're more reacty
* Refactor shared components like `Text` and `Header`, removing inline classNames that cannot be dependably overridden by passing in new classNames. (like we did with `ActionButton`)

PR #257, opened 2021-05-27
#124 Fix CSS Classes & Refactor Button

### LINKED ISSUE #124: get rid of dynamic naming classes
when we use dynamic classes names we lose the ability of purge non-used css classes from tailwind

I added this TODO a while back: https://github.com/eoscommunity/Eden/blob/main/packages/webapp/tailwind.config.js#L2

Because of this: 
https://github.com/eoscommunity/Eden/blob/cb8515d/packages/webapp/src/_app/ui/button.tsx#L33-L52

And we also would need to fix this little one:
https://github.com/eoscommunity/Eden/pull/120/files#r626227720

sparkplug0025:
- Fixes #124, enabling Tailwind purge
- First stab at #251


cool-ant (approved):
Wow, lots to take in in the first few files, and then the pleasure of re-reading the refactored *consumer* pages is great! Looks great Sparky

cool-ant (approved):

PR #258, opened 2021-05-27
Automate closing finished auctions as part of gc

swatanabe:
- If there are no bids, the assets will be burned
- A new migrate action sets up the tables required to track auctions that started before the new contract was deployed.
- The 5% collection fee is not claimed

### DIFF
--- a/contracts/eden/src/auctions.cpp
+++ b/contracts/eden/src/auctions.cpp
@@ -44,15 +44,16 @@ namespace eden
       using auction_table_type = eosio::multi_index<
           "auctions"_n,
           auction,
-          eosio::indexed_by<"byassets"_n,
-                            eosio::const_mem_fun<auction, eosio::checksum256, &auction::by_assets>>>;
+          eosio::indexed_by<
+              "byassets"_n,
+              eosio::const_mem_fun<auction, eosio::checksum256, &auction::by_assets>>>;
    }  // namespace atomicmarket
 
    void auctions::add_auction(uint64_t asset_id)
    {
       auction_tb.emplace(contract, [&](auto& row) {
-         row.asset_id = asset_id;
-         row.last_known_end_time =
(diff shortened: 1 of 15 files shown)


tbfleming:
Nice migration system!

contracts/eden/src/auctions.cpp
```diff
@@ -0,0 +1,180 @@
...
+         // We can distinguish the right auction by seller.
+         auto key = eosio::sha256(reinterpret_cast<const char*>(&auction.asset_id()),
+                                  sizeof(auction.asset_id()));
+         auto market_iter = asset_id_idx.lower_bound(key);
+         while (true)
+         {
+            if (market_iter->asset_ids.size() != 1 ||
+                market_iter->asset_ids.front() != auction.asset_id())
+            {
+               market_iter = asset_id_idx.end();
```
tbfleming, line 99: This assumes there won't be a hash collision. Probably ok because:
* It will be astronomically rare
* If it happens, the result isn't a big deal

tbfleming (approved):

contracts/eden/src/auctions.cpp
swatanabe: Also, any hash collision means that the security of sha256 is compromised, and we have much bigger problems.

tbfleming: One of the eosio chains, I forgot which, had a collision between 2 different transactions. They were far enough apart that nodeos let them through. Several history solutions fell over and had to get workarounds.

swatanabe: Were they deferred transactions?

swatanabe: Identical transactions will have the same hash, and deferred transactions have tapos cleared.

tbfleming: I don't remember

Issue #253, opened 2021-05-26
🎨 Induction cancellation improvements


brandonfancher:
* An inviter may also want to cancel a non-expired induction if one of the witnesses is not available or not responsive. Or if they're an initial Genesis member and have reached their 15-induction limit.
* There may be a use case for an invitee or witness to cancel too. We should figure out how to expose that.
* It'd be good if this error message tells the user they should cancel the existing on first and try again.
![image](https://user-images.githubusercontent.com/7911424/119689347-c47cc000-be16-11eb-9858-940813ac6850.png)

_Originally posted by @brandonfancher in https://github.com/eoscommunity/Eden/issues/236#issuecomment-848876935_

brandonfancher:
Recommendation @sparkplug0025: Instead of the model we have now where the "Expired" button _is_ the cancel button...what if we do something like this:

![image](https://user-images.githubusercontent.com/7911424/119737210-ea24bc00-be4c-11eb-8f40-168a793bbd6e.png)

... and on hover:
![image](https://user-images.githubusercontent.com/7911424/119737241-f3ae2400-be4c-11eb-8588-41a603c0301e.png)

We could do this next to any status that is deletable (which I think is all or most, right?) And in that way, the inviter, a witness, or the invitee can cancel/revoke the invite at any point.

I actually ran into this problem myself just an hour ago! I had an induction scheduled today with an invitee who had a very small window of availability... and one of the witnesses didn't show up. We were able to get a replacement witness, but we had to make the other witness the new invitee...since I couldn't reinvite him myself.  So this is already becoming a problem.

brandonfancher:
If you like it, go for it!

PR #260, opened 2021-05-28
#253 Cancellation Improvements

sparkplug0025:
Implements #253

### DIFF
--- a/packages/webapp/src/_app/ui/button.tsx
+++ b/packages/webapp/src/_app/ui/button.tsx
@@ -71,7 +71,7 @@ export const Button = ({
     target,
     isExternal,
 }: Props) => {
-    const baseClass = "inline-block border focus:outline-none";
+    const baseClass = "inline-block border focus:outline-none text-center";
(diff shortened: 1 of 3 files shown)


sparkplug0025:
I saw this and I actually thought it was a good behavior since we might have the witness names in the middle. I will take a better look into this over the weekend.

brandonfancher:
> I saw this and I actually thought it was a good behavior since we might have the witness names in the middle. I will take a better look into this over the weekend.

Yeah, not the end of the world if we need it to wrap because of space constraints. We should prob center the text in the button horizontally when that happens though.

brandonfancher:
> > I saw this and I actually thought it was a good behavior since we might have the witness names in the middle. I will take a better look into this over the weekend.
> 
> Yeah, not the end of the world if we need it to wrap because of space constraints. We should prob center the text in the button horizontally when that happens though.

Prob as simple as reintroducing the `text-center` I removed in that last PR.

PR #264, opened 2021-05-30
add witnesses to the waiting endorsements table

sparkplug0025:
- [ ] ~~can I refactor the witnesses name fetching from `inviter-inductions.tsx` + `endorser-inductions.tsx`?~~


sparkplug0025:
I'm going to actually flag the above refactor as a new issue since I also noticed we have similar logic on `invitee-inductions.tsx` as well.

PR #269, opened 2021-06-02
Global Pending Inductions #247

### LINKED ISSUE #247: 🎨 Global list of currently pending invitations and their status
You can easily a list of invitations you're a party to, but you cannot see other's invitations without using a block explorer. These are common questions: Who's been invited? How many invites are out there?

Expose a version of our table component somewhere that shows all pending inductions...paginated, of course.

sparkplug0025:
Fix #247 

image

### DIFF
--- a/packages/webapp/src/pages/induction/pending-invitations.tsx
+++ b/packages/webapp/src/pages/induction/pending-invitations.tsx
@@ -3,6 +3,7 @@ import { useQuery } from "react-query";
 import { SingleColLayout, Card, PaginationNav, membersStatsQuery } from "_app";
 import { getInductions } from "inductions";
 import { useEffect, useState } from "react";
+import { SpectatorInductions } from "inductions";
 
 const QUERY_INDUCTIONS = "query_inductions";
 const PAGE_SIZE = 10;
@@ -46,28 +47,18 @@ export const PendingInvitationsPage = () => {
 
     return (
         <SingleColLayout>
-            <Card title="Pending Inductions" titleSize={2}>
-                {inductions.isLoading && "Loading pending inductions..."}
-                {inductions.error && "Fail to load inductions"}
-                {inductions.data && (
-                    <>
-                        <ul>
-                            {inductions.data.map((i) => (
-                                <li key={i.id}>
-                                    {i.inviter} - {i.invitee}
-                                </li>
-                            ))}
-                        </ul>
-                        <PaginationNav
-                            paginate={paginateInductions}
-                            hasNext={inductions.data.length >= PAGE_SIZE}
-                            hasPrevious={
-                                inductions.data[0].id !== page.firstKey
-                            }
-                        />
-                    </>
-                )}
-            </Card>
+            {inductions.isLoading && "Loading pending inductions..."}
+            {inductions.error && "Fail to load inductions"}
+            {inductions.data && (
+                <div className="space-y-4">
+                    <SpectatorInductions inductions={inductions.data} />
+                    <PaginationNav
+                        paginate={paginateInductions}
+                        hasNext={inductions.data.length >= PAGE_SIZE}
+                        hasPrevious={inductions.data[0].id !== page.firstKey}
+                    />
+                </div>
+            )}
         </SingleColLayout>
     );
 };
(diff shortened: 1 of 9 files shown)


brandonfancher:
This is great. I'll do a more full review soon--if not this weekend, then Monday. A few thoughts:

1. I think most status buttons should be `neutral` unless it's an actionable status for an induction you are a party to. For example, in the screenshot below, "Complete ceremony" looks like I should be able to take action to complete it...but I'm not a party to that induction so that button should probably look like the other buttons and say "Waiting for ceremony" or something like that.
image

2. I have a different entry point to this page in mind. Instead of going to Community and seeing the message there under new members, I envision a link under the Call to Action (or your current invitations, if you have any) saying something like: "View all pending invitations to Eden." But I don't have that well-defined yet. We should prob merge what's here (after my review and changes to buttons) and then I can tweak that. I also see the "N pending invitations" stat on the home page linking here too.

sparkplug0025:
> I think most status buttons should be neutral unless it's an actionable status for an induction you are a party to. For example, in the screenshot below, "Complete ceremony" looks like I should be able to take action to complete it...but I'm not a party to that induction so that button should probably look like the other buttons and say "Waiting for ceremony" or something like that.

Nice catch!

sparkplug0025:
> I have a different entry point to this page in mind. Instead of going to Community and seeing the message there under new members, I envision a link under the Call to Action (or your current invitations, if you have any) saying something like: "View all pending invitations to Eden." But I don't have that well-defined yet. We should prob merge what's here (after my review and changes to buttons) and then I can tweak that. I also see the "N pending invitations" stat on the home page linking here too.

I like both ideas. Let's merge it for now and iterate on it later.

packages/webapp/src/pages/induction/pending-invitations.tsx
```diff
@@ -0,0 +1,66 @@
+import { useQuery } from "react-query";
+
+import { SingleColLayout, Card, PaginationNav, membersStatsQuery } from "_app";
+import { getInductions } from "inductions";
+import { useEffect, useState } from "react";
+import { SpectatorInductions } from "inductions";
+
+const QUERY_INDUCTIONS = "query_inductions";
+const PAGE_SIZE = 10;
```
brandonfancher, line 9: Somethings off with the math around this--an off-by-one somewhere. If there are 10 pending invitations, the "Next Page" button still shows up. 11, it doesn't.

For example, I set this to `5` in my dev environment (there are 5 pending invitations globally) and the "Next Page" button shows up.

image

brandonfancher: It'd be worth checking this elsewhere we're paginating too. I *think* I've seen this problem in the Community pagination too.

packages/webapp/src/pages/induction/pending-invitations.tsx
```diff
@@ -0,0 +1,66 @@
+import { useQuery } from "react-query";
+
+import { SingleColLayout, Card, PaginationNav, membersStatsQuery } from "_app";
```
brandonfancher, line 3: We can get rid of `Card` and `membersStatsQuery` here.

brandonfancher (approved):
A couple comments to address, but once addressed, merge away!

packages/webapp/src/pages/induction/pending-invitations.tsx
swatanabe: Like this: https://github.com/eoscommunity/Eden/blob/main/packages/webapp/src/pages/members/index.tsx#L107

swatanabe: In order to know whether there are more entries, you would need to ask for one more than you can display.

sparkplug0025: exactly. It was a conscious decision to leave this next button there to save one extra request. Do you want me to fix by adding this extra request?

brandonfancher: Yeah, let’s make the change. It’s not an entirely new request, though, right? Can we just fetch pageSize + 1 per page?

brandonfancher: (Or maybe it is if we have to fetch each record individually...) I’m not in front of the code atm.

brandonfancher: Either way, worth it.

PR #271, opened 2021-06-03
Track donated funds

swatanabe:
- Direct token transfers out of the contract are no longer allowed.  Transfers using donated funds must use the transfer action.
- There is now a plain `donate` action.

### DIFF
--- a/contracts/eden/include/accounts.hpp
+++ b/contracts/eden/include/accounts.hpp
@@ -8,6 +8,16 @@
 
 namespace eden
 {
+   struct migrate_account_v0
+   {
+      eosio::name last_visited;
+      // user_total = \sum balance(account) \forall account <= last_visited
+      eosio::asset user_total;
+      uint32_t migrate_some(eosio::name contract, uint32_t max_steps);
+      void adjust_balance(eosio::name owner, eosio::asset amount);
+   };
+   EOSIO_REFLECT(migrate_account_v0, last_visited, user_total)
+
    struct account_v0
    {
       eosio::name owner;
@@ -19,6 +29,13 @@ namespace eden
 
    using account_variant = std::variant<account_v0>;
 
+   // invariant: The total of all accounts (including internal and user accounts)
+   // is equal to the eosio.token balance of the contract.
+   //
+   // scopes:
+   // - default_scope: users
+   // - "owned": internal accounting
+   // - "outgoing": Should never exist outside a transaction.
    struct account
    {
       account_variant value;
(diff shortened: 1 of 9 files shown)


contracts/eden/src/actions/induct.cpp
```diff
@@ -81,12 +82,23 @@ namespace eden
...
+      accounts user_accounts{get_self()};
+      accounts internal_accounts{get_self(), "owned"_n};
 
       const auto& induction = inductions.get_induction(id);
       eosio::check(payer == induction.invitee(), "only inductee may donate using this action");
       eosio::check(quantity == globals.get().minimum_donation, "incorrect donation");
-      accounts.sub_balance(payer, quantity);
+      user_accounts.sub_balance(payer, quantity);
+      migrations migrations{get_self()};
+      if (auto migration = migrations.get<migrate_account_v0>())
```
tbfleming, line 93: It'd be nice to roll migration handling into add_balance and sub_balance, or create a higher level of add_balance and sub_balance

swatanabe: The main difficulty is that each use case (donation, deposit, and withdrawal) behaves differently.

swatanabe: Also, the set of actions that need to handle migration is fixed, as actions yet to be added can just assert that this migration is complete.

contracts/eden/include/accounts.hpp
```diff
@@ -8,6 +8,16 @@
 
 namespace eden
 {
+   struct migrate_account_v0
+   {
+      eosio::name last_visited;
```
tbfleming, line 13: Uninitialized fields

contracts/eden/include/migrations.hpp
```diff
@@ -14,19 +16,42 @@ namespace eden
       uint32_t migrate_some(eosio::name contract, uint32_t max_steps) { return max_steps; }
    };
    EOSIO_REFLECT(no_migration<0>);
-   using migration_variant = std::variant<migrate_auction_v0, no_migration<0>>;
+   using migration_variant = std::variant<migrate_auction_v0, migrate_account_v0, no_migration<0>>;
```
tbfleming, line 19: If the contract before this PR is currently at index 1, then it looks like deserialization will malfunction when this PR is installed. Did you mean this instead?

```
using migration_variant = std::variant<migrate_auction_v0, no_migration<0>, migrate_account_v0, no_migration<1>>;
```

contracts/eden/include/accounts.hpp
tbfleming: Never mind. I somehow forgot about those types' constructors.

contracts/eden/include/migrations.hpp
swatanabe: I intended to do that iff we merge to production before this PR is merged.

tbfleming (approved):

PR #272, opened 2021-06-04
Queries Refactor

sparkplug0025:
This PR is preparing the terrain for #243 , forcing the use of queries across the board and removing arbitrary ones.

What it does:
- centralizes query
- no more arbitrary queries, we reuse the ones defined in the `queries.ts`
- removes useFetchedData hook, now we only fetch through queries
- a little refactor in the inductions container page

PR #273, opened 2021-06-08
Refactor status button and Inductions table data

sparkplug0025:
Follow up to issue identified on #272 and #85 - [specifically here](https://github.com/eoscommunity/Eden/pull/272#issuecomment-856705020) 

• Look at how simple the inductions table are now (specifically invitee and inviter -- endorsers requires a little bit of stretch just to fetch the inductions, but the hook was refactored out of the loop.


packages/webapp/src/inductions/components/induction-lists/invitee-inductions.tsx
```diff
@@ -47,39 +42,23 @@ const INVITEE_INDUCTION_COLUMNS: InductionTable.Column[] = [
 
 const getTableData = (inductions: Induction[]): InductionTable.Row[] => {
     return inductions.map((induction) => {
-        const { data: allEndorsements } = useQuery(
-            queryEndorsementsByInductionId(induction.id)
-        );
-
         const { data: inviter } = useMemberByAccountName(induction.inviter);
```
brandonfancher, line 45: This hook is still inside the map. Are we wanting to get rid of that one too?

brandonfancher (approved):
Love it. Tested it extensively. All good. Just the one question about the remaining hook inside the loop/map. Other than that, approved. ✅

packages/webapp/src/inductions/components/induction-lists/invitee-inductions.tsx
sparkplug0025: nice catch! Working on it!

PR #280, opened 2021-06-14
Zoom Auth + Meeting Creation POC

sparkplug0025:
Adds a POC of authenticating and refreshing the token with Zoom.

image


brandonfancher:
@sparkplug0025 I tested this with our app client_id and secret. Works great!

1. I noticed the meeting is set to record automatically, locally. Nice!
2. Do you know if there's any option for us to disable the waiting room so that the host doesn't have to manually let people in?

image

packages/webapp/src/config.ts
```diff
@@ -123,3 +127,9 @@ export const jobKeys = {
 export const eosPrivateKeys = {
     gcJob: process.env.EOS_PRIVATE_KEY_GC_JOB || "",
 };
+
+export const zoom = {
+    clientKey: process.env.NEXT_PUBLIC_ZOOM_CLIENT_ID || "",
+    clientSecret: process.env.ZOOM_CLIENT_SECRET || "",
+    oauthRedirect: `http://localhost:3000/oauth/zoom`, // todo: make the domain dynamic
```
cool-ant, line 134: Does your plugin catch "todo" lowercase? The one I use i believe require it to be "TODO" to be picked up. I just don't want this to be lost.

packages/webapp/src/pages/api/meeting-links.ts
```diff
@@ -0,0 +1,75 @@
...
+        const meeting = await generateMeeting(result.data);
+        return res.status(200).json({ meeting });
+    } catch (error) {
+        console.error(error);
+        return handleErrors(res, error);
+    }
+};
+
+export const generateMeeting = async (meetingRequest: MeetingLinkRequest) => {
+    console.info(meetingRequest);
```
cool-ant, line 38: spurious?

packages/webapp/src/pages/api/refresh-zoom.ts
```diff
@@ -0,0 +1,40 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import * as z from "zod";
```
cool-ant, line 2: very minor update: you can do this as `import { z } from "zod"`

packages/webapp/src/pages/oauth/zoom.tsx
```diff
@@ -0,0 +1,145 @@
...
+                    </p>
+                </div>
+            ) : (
+                <Button href={zoomConnectAccountLink}>
+                    Link your Zoom Account
+                </Button>
+            )}
+        </div>
+    );
+};
```
cool-ant, line 145: There's a few console calls here. I'm find with them, but wanted to point them out in case they are trimmable now.

cool-ant (approved):
very minor things

packages/webapp/src/pages/oauth/zoom.tsx
sparkplug0025: thanks! trimmed the less important ones!

packages/webapp/src/config.ts
sparkplug0025: changed to uppercase. I actually don't have a plugin. I do a glob search for `todo:`. But good to know, I will use that one! :D

sparkplug0025:
> @sparkplug0025 I tested this with our app client_id and secret. Works great!
> 
> 1. I noticed the meeting is set to record automatically, locally. Nice!
> 2. Do you know if there's any option for us to disable the waiting room so that the host doesn't have to manually let people in?
> 
> These were the options set on that meeting:
> 
> image

I removed the join before host and changed the waiting room to false!

brandonfancher:
yeah, just to recap, we want to make sure that the settings allow for anyone to join the room at any time. So I believe "Waiting Room" should be false and "Allow participants to join at any time" should be true (I'm referring to the UI settings above...of course, the underlying configuration flags might be worded differently.)

PR #319, opened 2021-06-22
POC: Local IPFS Node Upload

sparkplug0025:
It works for images but takes forever for files like 500mb videos... Even for 100mbs it takes a while :(


brandonfancher:
Closing since we went a different direction.

PR #320, opened 2021-06-23
Init Box

sparkplug0025:
Bootstraps Eden Box server.
- add a new box package
- sets up a basic express
- sets up logging mechanism
- dockerize
- adds CI


brandonfancher (approved):
📦

Issue #284, opened 2021-06-16
eden: add community treasury balance to home page stats


brandonfancher:
Depends on #283. We should add this balance initially to the homepage. Later, as we fill in more details around election representative disbursements, this will be its own Treasury page with more details. But that comes later.

Ship it.

PR #321, opened 2021-06-24
#284 Add treasury balance to Community Stats page

cool-ant:


packages/webapp/src/_app/utils/asset.ts
```diff
@@ -5,6 +5,14 @@ export interface Asset {
...
 
 export const assetFromString = (asset: string): Asset => {
+    const re = /^[0-9]+\.[0-9]+ [A-Z,a-z]{1,5}$/
+    if(asset.match(re) === null) {
+        return {
+            symbol: "ERR",
+            quantity: 0.0,
+            precision: 2
+        }
+    }
```
cool-ant, line 15: Verify the format to ensure we successfully parse it, and provide obvious feedback to the user in case of error

packages/webapp/src/_app/api/eden-contract.ts
```diff
@@ -1,7 +1,7 @@
 import { CONTRACT_GLOBAL_TABLE, getTableRows } from "_app";
 
 export const getIsCommunityActive = async (): Promise<boolean> => {
-    const rows = await getTableRows(CONTRACT_GLOBAL_TABLE, "community");
+    const rows = await getTableRows(CONTRACT_GLOBAL_TABLE, {lowerBound: "community"});
```
brandonfancher, line 4: Would you mind double checking your prettier auto-formatting? I believe there should be spaces inside of your curly braces here; autoformat would fix that.

packages/webapp/src/_app/eos/api.ts
```diff
@@ -16,17 +18,29 @@ interface TableResponse<T> {
...
+
 export const getRow = async <T>(
     table: string,
-    keyName?: string,
-    keyValue?: string
+    options?: TableQueryOptions,
 ): Promise<T | undefined> => {
-    const rows = await getTableRows(table, keyValue);
+    options = getTableParamsWithDefault(options)
+    const rows = await getTableRows(table, options)
```
brandonfancher, line 37: Here and below on Line 55, I'd prob just do a spread for options: `{ ...TableParamDefaults, ...options }` and do away with the `getTableParamsWithDefault` function altogether. And because the `TableParamDefaults` is a constant, I'd do all caps: `TABLE_PARAM_DEFAULTS` or whatever name makes the most sense.

const rows = await getTableRows

packages/webapp/src/_app/utils/asset.ts
brandonfancher: I like that you're adding validation here.

As for what to do when the validation fails, I think I'd be more comfortable `throw`ing an error here rather than returning something to the UI/view layer that acts like everything's okay. It should be the job of the UI layer to decide what to display to the user in the event of an exception.

brandonfancher: ...that said, I'm okay with the idea of doing that in a follow-up PR, since we'll need to handle that new exception in various other places in the code...and that might blow the scope of this one.

packages/webapp/src/members/api/eden-contract.ts
```diff
@@ -1,13 +1,21 @@
...
 
 export const getEdenMember = (account: string) =>
-    getRow<EdenMember>(CONTRACT_MEMBER_TABLE, "account", account);
+    getRow<EdenMember>(CONTRACT_MEMBER_TABLE, {keyName: "account", keyValue: account});
 
 export const getMembersStats = async () =>
     getRow<MemberStats>(CONTRACT_MEMBERSTATS_TABLE);
+
+export const getTreasuryStats = async () => {
+    const { balance } = await getRow<any>(CONTRACT_ACCOUNT_TABLE, {scope: "owned", keyName: "master"});
```
brandonfancher, line 19: yeah, this file needs some prettier autoformat too

packages/webapp/src/pages/index.tsx
```diff
@@ -82,38 +90,44 @@ export const Index = () => {
...
+    memberStats,
+    treasuryBalance,
+}: CommunityStatsProps) =>
+    memberStats && (
+        <Card className="flex flex-col justify-center items-center h-full space-y-4 text-md lg:text-lg">
+            <Heading size={2} className="mb-2">
+                Community Stats
+            </Heading>
+            <Text className="font-medium" size="md">
+                {`Treasury: ${assetToString(treasuryBalance, 4)}`}
```
brandonfancher, line 120: I think this could be simplified to: `Treasury: {assetToString(treasuryBalance, 4)}`

brandonfancher (changes requested):
This is really good. A couple minor comments after my review of the code.

packages/webapp/src/_app/utils/asset.ts
cool-ant: @brandonfancher I'd propose just doing it now, given this is a case we're already not handling in any way. Might as well do this part right and run into what we're going to run into anyway and have this part right. Lmk if you disagree; otherwise, that'll be my plan.

packages/webapp/src/_app/eos/api.ts
cool-ant: @brandonfancher That's what I tried to do originally, but I found that the override order didn't match what I expected... I'll recheck those findings though.

brandonfancher:
@cool-ant I spun it up locally and logged in as an existing user. The membership page no longer recognizes the logged-in member. For example, in the screenshot below, pending outgoing invitations should be showing up and the username in the right side of the nav bar should read "Eden Member 11" and not "edenmember11".

image

vs.

image

packages/webapp/src/_app/eos/api.ts
cool-ant: hm, seems to work. Apparently too much <joke drug of choice> last night? :)

packages/webapp/src/_app/utils/asset.ts
brandonfancher: Adding the throw in there is fine. But we'll also want to wrap any invocation of this function in a `try...catch` too. Otherwise, we may run into unhandled exceptions.

cool-ant:
> @cool-ant I spun it up locally and logged in as an existing user. The membership page no longer recognizes the logged-in member. For example, in the screenshot below, pending outgoing invitations should be showing up and the username in the right side of the nav bar should read "Eden Member 11" and not "edenmember11".
> 
> image
> 
> vs.
> 
> image

Interesting. Good catch. It was a subtlety of the original code, which hopefully is now explicit how that was being performed...

packages/webapp/src/_app/eos/api.ts
```diff
@@ -30,6 +30,7 @@ export const getRow = async <T>(
     options?: TableQueryOptions
 ): Promise<T | undefined> => {
     options = { ...TABLE_PARAM_DEFAULTS, ...options };
+    options.lowerBound = options.keyValue;
```
brandonfancher, line 33: Another option: `options = { ...TABLE_PARAM_DEFAULTS, ...options, lowerBound: options?.keyValue };`
Just a comment...either way technically works.

That said, something feels weird about replacing `lowerBound` here with `keyValue`. For example, any time `getTableRows` is called via `getRow`, `lowerBound` (if passed in) will be discarded.

I didn't write this portion of the code in the first place; it'd be good for us to wait for @sparkplug0025 to weigh in before merging. Otherwise, I can do a deeper dive and give it some more thought tomorrow.

I think that before, `getRow` and `getTableRows` had distinct APIs (param types). Now, with the introduction of the `TableQueryOptions` interface, they're converging. And it seems we have two keys (`lowerBound` and `keyValue`) that basically do the same thing. We should probably try to avoid that.

brandonfancher:
Now that `assetToString()` throws, we should `try...catch` anywhere we consume it. (That's why I suggested we could introduce the `throw` in a follow-up PR.)

packages/webapp/src/_app/eos/api.ts
cool-ant: Well, lowerBound is just that, and what we're saying is that we want the first record that matches the value. I'll certainly wait for @sparkplug0025 to weigh in, but this does emulate what was there before and I think just represents an odd use of bounds. This query *should* only ever return one record, but I'm assuming that by specifying the sought value as the lowerBound, we're ensuring that the first item with this value gets returned.

cool-ant: As to the assignment being part of the default value application, I felt they were 2 different operations that were easier to understand when kept separate, especially given the esoteric oddity of the 2nd line. I didn't want that all muddled in the same line.

brandonfancher: `keyValue` and `keyName` are only consumed (as convenience properties) by `getRow(...)`. They're not actually consumed by `getTableRows(...)` directly...that's what's throwing me off. What if we separated it out in the interface. So `TableQueryOptions` would just be:

```
export interface TableQueryOptions {
    scope?: string;
    lowerBound?: any;
    upperBound?: any;
    limit?: number;
}
```

and the type on `getRow` for `options`:

```
export const getRow = async <T>(
    table: string,
    options?: { keyName?: string; keyValue?: string; } & TableQueryOptions,
): Promise<T | undefined> => {
```

That way, it's clear that `keyName` and `keyValue` are only for `getRow()` and not `getTableRows()`?

brandonfancher (approved):

packages/webapp/src/members/api/eden-contract.ts
```diff
@@ -1,13 +1,28 @@
...
+    getRow<EdenMember>(CONTRACT_MEMBER_TABLE, {
+        keyName: "account",
+        keyValue: account,
+    });
 
 export const getMembersStats = async () =>
     getRow<MemberStats>(CONTRACT_MEMBERSTATS_TABLE);
+
+export const getTreasuryStats = async () => {
+    const { balance } = await getRow<any>(CONTRACT_ACCOUNT_TABLE, {
```
sparkplug0025, line 22: you should probably create an interface for the TreasuryStats (see eden member above), so we can get rid of any

packages/webapp/src/_app/eos/api.ts
```diff
@@ -35,22 +51,21 @@ export const getRow = async <T>(
...
         key_type: "",
-        limit: `${limit}`,
-        lower_bound: lowerBound,
-        upper_bound: upperBound,
+        limit: `${options.limit}`,
+        lower_bound: options.lowerBound,
+        upper_bound: options.upperBound,
         reverse,
-        scope: CONTRACT_SCOPE,
+        scope: options.scope,
```
sparkplug0025, line 68: nice refactor, I love the TableQueryOptions idea!

packages/webapp/src/_app/eos/api.ts
sparkplug0025: The way that I see this is we have two functions:
1. `getRow`: to get a single record based on the key (like a pretty simple get from a kv store)
2. `getRows`: to effectively setup, execute a query and work on the results.

That being said I'd probably remove the `TableQueryOptions` altogether from `getRow` because it just does not make sense to set these query options in a simple get. Then I'd rollback and keep using the original interface `getRow(table, keyName, keyValue)`.

Also, on top of that we could rename `getRows` to `queryEosTable` so we don't make any confusion between them.

`getRow` needs to fetch the first record from the `lowerbound` results (returned by `getRows`) and see if it matches with the expected key (hence the limit 1 and keyName comparison).

packages/webapp/src/pages/interfaces.ts
```diff
@@ -0,0 +1,4 @@
+export interface TreasuryStats {
```
sparkplug0025, line 1: this file can't exist in the `pages` folder. the `pages` folder is a convention for next, only routed pages should exist here, since you don't have any react component to render this "/interfaces" page it's breaking the build.

PR #322, opened 2021-06-24
Box IPFS Upload

sparkplug0025:
Sorry for the huge PR ☺️

- Adds ipfs upload endpoint to the box
- Adds a configuration on eden webapp to use the box or not
- Adds flexible cid field parsing from the action data
- Adds a `common` package to share APIs and common functionalities between the box and the webapp


packages/box/src/handlers/ipfs-upload.ts
```diff
@@ -0,0 +1,141 @@
...
+        );
+        if (parsedRequest.success !== true) {
+            throw new BadRequestError(parsedRequest.error.flatten());
+        }
+
+        const requestData: IpfsUploadRequest = parsedRequest.data;
+
+        if (!requestData.syncUpload) {
+            // We need to implement a background job mechanism to make sure the
+            // file can be upload asynchronously without blocking the client.
```
cool-ant, line 42: We should probably mark this obviously with a TODO so we can easily find this and be reminded to come back to it

packages/common/src/backend/error-handlers.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (error instanceof ServerError) {
+        return res
+            .status(error.status)
+            .json({ type: error.name, error: error.error });
+    } else {
+        return res
+            .status(500)
+            .json({ error: error.message || "Unknown error" });
+    }
+};
```
cool-ant, line 38: So clean... very nicely done

packages/common/src/eos/ipfs-actions.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    try {
+        // Extract the CID supporting dot notation fields, eg: new_member_profile.img
+        const cidFields = validAction.cidField.split(".");
+        let cid = data;
+        while (cidFields.length) {
+            const cidField = cidFields.shift()!;
+            cid = cid[cidField];
+        }
+        return cid;
```
cool-ant, line 75: I find this loop to be very hard to understand. Likely just that I don't know what the data looks like, but maybe you add a comment that explains or shows a tiny sample of data? I'm guessing cidFields is a dot-separated list of properties of an object, and this function is using cidFields to drill into the object to get the cid, which is nested? Some way of making that easier to get than re-reading this code would be great.

cool-ant (approved):
Just a couple minor comments. This all looks like super solid code. *Great* work.

packages/box/src/handlers/ipfs-upload.ts
sparkplug0025: done! 👍

packages/common/src/eos/ipfs-actions.ts
sparkplug0025: that's exactly it. just added the comment! done! 👍

brandonfancher:
@sparkplug0025 Progress. I've started reviewing this. Will finish tomorrow. I am noticing the build is failing (which you may already be aware of.) Sent you creds for a couple platforms for testing deployment of the docker containers too.

packages/box/src/handlers/ipfs-upload.ts
```diff
@@ -0,0 +1,141 @@
+import { Request, Response } from "express";
+import multer from "multer";
```
brandonfancher, line 2: TIL of multer. Nice!

brandonfancher (approved):
This is awesome. Approved.

I do have a suggestion: The IPFS upload behavior and the box vs. infura options are not going to be immediately clear to other devs (internally _or_ those who might want to set up Eden at a different domain/elsewhere). I recommend adding:

1. to the webapp README a note about the two IPFS upload options and how to use them ((Infura + NextJS Serverless Pin + Pinata) vs. (Box + Pinata)) including env vars;
2. to the box README a note about any env vars that need to be set up.
3. to any package READMEs a note about how to whitelist contracts/actions.

Basically, just some light documentation.

But again, this is really nicely done. 👏 👏 👏

sparkplug0025:
@brandonfancher great readme docs update suggestions! Working on it right now!

PR #329, opened 2021-07-01
Feature/fix social links

cool-ant:


packages/webapp/src/members/helpers/social-links.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (str && str.startsWith("@")) {
+        return str.slice(1);
+    }
+
+    return str;
+};
+
+export const getStringAfterTrailingSlash = (
+    str: string | undefined
+): string | undefined => str && str.slice(str.lastIndexOf("/") + 1);
```
brandonfancher, line 28: `https://www.linkedin.com/in/username/` is also a valid URL. We should trim off the last `/` if it's the last character and then pull out the username.

brandonfancher: Also, we should consider applying this `getStringAfterTrailingSlash()` to Twitter and Telegram too...before the `stripLeadingAtSymbol()` operation...just in case people past in links there too. I don't think it will hurt.

In fact, we could take it a step further and do both to all four in that order. 🤷‍♂️

brandonfancher (changes requested):

packages/webapp/src/members/helpers/social-links.ts
brandonfancher: If we do the same to all four, this could become one social handle prep function instead of four separate ones. Your call.

cool-ant: I agree. I've got a PR I'm almost ready for review on that has this change. Rather than porting that code to this branch, I'll just do that refactor there.

brandonfancher (approved):

PR #332, opened 2021-07-06
Bringing over the social link processing code

cool-ant:


packages/webapp/src/members/helpers/social-links.ts
```diff
@@ -1,28 +1,22 @@
...
 
-export const getStringAfterTrailingSlash = (
+const startsWithAtSymbol = (str: string | undefined) =>
+    str && str.startsWith("@");
+
+const stripLeadingAtSymbol = (str: string | undefined): string | undefined =>
+    str && startsWithAtSymbol(str) ? str.slice(1) : str;
+
+const containsSlashes = (str: string | undefined) =>
+    str && str.indexOf("/") >= 0;
```
brandonfancher, line 18: You could do a `str.includes("/")` here. More idiomatic. And, it's so readable, you could just inline it above and get rid of the extra function altogether. Just an idea.

packages/webapp/src/members/helpers/social-links.ts
```diff
@@ -1,28 +1,22 @@
...
-        return str.slice(1);
-    }
-
-    return str;
+    return stripLeadingAtSymbol(link);
 };
 
-export const getStringAfterTrailingSlash = (
+const startsWithAtSymbol = (str: string | undefined) =>
+    str && str.startsWith("@");
```
brandonfancher, line 12: This would be very readable inline too. Either way.

brandonfancher (approved):
LGTM

PR #333, opened 2021-07-07
Add (styled) react-modal component. Add xs breakpoint.

brandonfancher:
## About
* Uses `react-modal`
* Should be accessible, highly configurable
* Clicking on overlay/background and ESC key dismisses. Or we can disable either/both with a prop.

Screenshots below are for demonstration purposes only. Resign functionality is not part of this PR.

### Mobile
image
### Desktop
image


cool-ant (approved):

packages/webapp/src/_app/ui/modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+const mobileClass = "w-screen h-screen";
+const desktopClass = "md:h-auto md:rounded md:w-4/6 xl:w-6/12 2xl:w-2/5";
+
+interface Props {
+    title?: string;
+    children?: React.ReactNode;
+}
+
+/**
+ *
```
sparkplug0025, line 16: nit: whitespace

packages/webapp/src/pages/_app.tsx
```diff
@@ -24,6 +25,8 @@ Router.events.on("routeChangeError", () => NProgress.done());
 dayjs.extend(localizedFormat.default);
 dayjs.extend(relativeTime.default);
 
+Modal.setAppElement("#__next");
```
sparkplug0025, line 28: curious question: why `#__next`? is it arbitrary or is it a magic keyword for next or react-modal etc?

sparkplug0025 (approved):
love the small pr just adding the lib

packages/webapp/src/pages/_app.tsx
brandonfancher: `Modal.setAppElement()` helps with a11y, and is intended to target the DOM element representing the app mounting point. Next tags that in the DOM with the id `__next`.

PR #342, opened 2021-07-13
Add table to store encrypted links

swatanabe:


contracts/eden/src/encrypt.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace eden
+{
+   encrypt::encrypt(eosio::name contract, eosio::name scope)
+       : contract(contract), encrypted_data_tb{contract, scope.value}
+   {
+   }
+
+   void encrypt::set(uint64_t id, const std::vector<encrypted_key>& keys, const eosio::bytes& data)
+   {
+      encrypted_data_tb.emplace(contract, [&](auto& row) {
```
tbfleming, line 12: This doesn't allow replacing the data; e.g. when the meeting link went bad (very frequent with Zoom)

contracts/eden/include/eden.hpp
```diff
@@ -173,18 +184,21 @@ namespace eden
               invitee,
               witnesses,
               ricardian_contract(inductinit_ricardian)),
+       action(inductmeeting, id, keys, data),
```
tbfleming, line 187: A hors is a hors...

contracts/eden/src/actions/elect.cpp
```diff
@@ -74,6 +75,20 @@ namespace eden
...
+   void eden::electmeeting(eosio::name account,
+                           uint8_t round,
+                           const std::vector<encrypted_key>& keys,
+                           const eosio::bytes& data)
+   {
+      eosio::require_auth(account);
+      members members{get_self()};
+      elections elections{get_self()};
+      auto group_id = elections.get_group_id(account, round);
+      members.check_keys(elections.get_group_members(group_id), keys);
```
tbfleming, line 87: Since a group member without a key stops the group from using this feature, should we require members set keys to be eligible to vote?

swatanabe: I don't think that use of this feature should be mandatory.  It's only a convenience.  If it's enforced at all, it should be enforced by the front end, not the contract.

tbfleming (approved):

PR #346, opened 2021-07-15
UI: MemberChip Abstraction and Permutations for Elections

brandonfancher:
Creates a `GenericMemberChip` component that is used by three new specific member chip components/permutations:

* `MemberChip`
* `VotingMemberChip`
* `WinningMemberChip`

These can be seen in action on the members pages and in a new page: `/election/components`.

image

image

### DIFF
--- a/packages/webapp/src/_app/ui/generic-member-chip.tsx
+++ b/packages/webapp/src/_app/ui/generic-member-chip.tsx
@@ -1,0 +1,53 @@
+import { ipfsUrl } from "_app";
+import { MemberData } from "members/interfaces";
+
+interface Props {
+    member: MemberData;
+    actionComponent?: React.ReactNode;
+    contentComponent: React.ReactNode;
+    onClickChip?: (e: React.MouseEvent) => void;
+    onClickProfileImage?: (e: React.MouseEvent) => void;
+}
+
+export const GenericMemberChip = ({
+    member,
+    onClickChip,
+    onClickProfileImage,
+    actionComponent,
+    contentComponent,
+}: Props) => (
+    <div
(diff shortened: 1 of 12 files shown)


packages/webapp/src/_app/ui/generic-member-chip.tsx
```diff
@@ -0,0 +1,53 @@
...
+    >
+        <div className="flex space-x-2.5">
+            <MemberImage
+                member={member}
+                onClick={onClickProfileImage || onClickChip}
+            />
+            {contentComponent}
+        </div>
+        {actionComponent}
+    </div>
```
cool-ant, line 32: Wow. Love this organization

packages/webapp/src/members/components/member-chip.tsx
```diff
@@ -0,0 +1,146 @@
...
+                />
+                <p className="text-sm tracking-tight leading-none p-t-px group-hover:underline">
+                    {assetToString(member.auctionData.price, 2)} (#
+                    {member.assetData?.templateMint})
+                </p>
+            </div>
+        );
+    }
+
+    return <></>;
```
cool-ant, line 121: Just a preference, but I'm a fan of having the intended code be unwrapped/primary. So
if (!member.auctionData) return <></>
//rest of code unrwapped

cool-ant (approved):

packages/webapp/src/members/components/member-chip.tsx
brandonfancher: Actually, that's usually my preference too. And less nesting. 👍

brandonfancher: Made that adjustment. Thanks!

brandonfancher:
A couple more UI tweaks: 
image

PR #347, opened 2021-07-16
Creating some initial data fetching and fixtures to fill in some of the data that we don't have yet in the contract

cool-ant:
Fixtures are done super simple with a .env var rather than trying to use nock.js or something similar. This is aimed at being simple and potentially going away. But we need something that works now and is flexible to dev and test a process that is hard to manually manipulate at the moment.

add to your .env.local to pick up the new fixtures
NEXT_PUBLIC_DEV_USE_FIXTURE_DATA = "true"


packages/webapp/src/delegates/api/fixtures.ts
```diff
@@ -0,0 +1,115 @@
...
+        },
+        status: 1, //active
+        election_participation_status: 0, // no_donation is kinda the default if other special cases don't apply
+        election_rank: 1,
+        representative: "edenmember11", // "parent" field
+    },
+    {
+        templateId: 84432,
+        createdAt: 1621635329000,
+        name: "Eden Member 11",
```
brandonfancher, line 36: So in this case, `Eden Member 11` is the Head Chief, right?

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ElectionState } from "elections/interfaces";
+import {
+    CONTRACT_CURRENT_ELECTION_TABLE,
+    CONTRACT_ELECTION_STATE_TABLE,
+    getTableRows,
+} from "_app";
+import { fixtureElectionState } from "./fixtures";
+
+export const getCurrentElection = async () => {
+    const rows = await getTableRows<any>(CONTRACT_CURRENT_ELECTION_TABLE);
```
brandonfancher, line 11: Do you also want a `if (devUseFixtureData) return fixtureCurrentElection;` here?

packages/webapp/src/nfts/api/nfts.ts
```diff
@@ -6,7 +6,7 @@ export const getTemplate = async (templateId: string) => {
     return templates.length ? templates[0] : undefined;
 };
 
-const LAUNCH_TIMESTAMP = "&after=1619779033000";
+const LAUNCH_TIMESTAMP = "&after=1626281073000";
```
brandonfancher, line 9: We shouldn't check this in.

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -19,8 +19,83 @@ interface Props {
...
+        data: memberStats,
+    } = useQuery({
+        ...queryMemberStats,
+        keepPreviousData: true,
+    });
+
+    return (
+        <RawLayout title="Election">
+            <Text size="sm" className="mb-8">
+                Note: Data is in square brackets if it's not Jason (to show if
```
brandonfancher, line 41: JSON

packages/webapp/src/pages/election/index.tsx
```diff
@@ -20,7 +28,125 @@ interface Props {
...
+    const electionStartDateTime =
+        currentElection &&
+        ((currentElection.election_seeder &&
+            currentElection.election_seeder.end_time) ||
+            currentElection.start_time);
+
+    return (
+        <RawLayout title="Election">
+            <Text size="sm" className="mb-8">
+                Note: Data is in square brackets if it's not Jason (to show if
```
brandonfancher, line 72: JSON?

brandonfancher (changes requested):

packages/webapp/src/delegates/api/fixtures.ts
cool-ant: right

packages/webapp/src/elections/api/eden-contract.ts
cool-ant: Yes, just didn't get to that yet; added a TODO to indicate that

packages/webapp/src/pages/delegates/index.tsx
cool-ant: lol

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -27,7 +52,7 @@ export const queryTreasuryStats = {
 
 export const queryMembersStats = {
     queryKey: "query_member_stats",
-    queryFn: getMembersStats,
+    queryFn: getMemberStats,
 };
```
brandonfancher, line 56: There are two queries exposed here now: `queryMemberStats` and `queryMembersStats`. They both have the same `queryKey` and `queryFn`. And when I tested in non-dev-fixtures mode, the member stats are not loading on the home page.

brandonfancher (approved):
LGTM!

PR #351, opened 2021-07-19
jbh without waiting room

sparkplug0025:


brandonfancher:
Documenting this here, while the code clearly has `waiting_room` set to `false`, it's still creating meetings for me with a Waiting Room. When I tested this with Thomas, I grabbed my meeting link, signed out of zoom, and joined the meeting link. Even though I wasn't signed in, it set me as the host (probably because I was the first one to join the meeting and "join before host" is enabled.) Then I had Thomas join, and it put him in the waiting room and asked me to admit him.

This is definitely better than not allowing anyone to join before the rightful host, but not as optimal as just letting everyone join whenever without waiting room.

As I dug into this, I found: https://devforum.zoom.us/t/meeting-creation-with-waiting-room-disabled-using-zoom-api/52801/2

Basically, it says that Zoom not forces all meetings to be secured with one of the following three settings: passcode, only-authenticated-users-can-join, or waiting room. If none of these are set, all meetings created will default to Waiting Room. And when I logged in to check my own settings, it says the same thing in my account UI.

So...I think the solution is to set a passcode during creation that's complex enough that it should satisfy most passcode requirements set by account holders or their admins. Then we ensure that the passcode is appended to the meeting link before we encrypt the link. If we do that, I believe Zoom will respect our `waiting_room: false` flag.

That's optimal, because it means that other attendees will not be stuck if something happens to the de facto host.

sparkplug0025:
@brandonfancher nice findings! can you test it again now that we have a random password?

brandonfancher:
It wasn't successfully setting the passcode on the meeting...thus the waiting room was still on. I looked at the docs and turns out the key is `password` (not `passcode`) and that it's not a child of `settings` but a _sibling_. Fixed that (see my commit) and it works! No more waiting room. Tested successfully with others.

brandonfancher:
So, this appears to fix the issue at hand (gets rid of the waiting room.) After testing more, we found that only the host (account under which the meeting was created) is able to record. The host can grant others permission to record locally...or they can make others hosts or co-hosts, and they can record that way. But even if others are granted host/co-host, if they choose the _cloud_ recording option, the recording saves to the Host's account...not the co-hosts account. That could get confusing.

If we want, we can go ahead and merge this. But we should investigate whether we can set these up so that all attendees have the ability to record locally without action required on the part of the original host.

This isn't a deal breaker, but fixing it will prevent some confusion for sure.

brandonfancher (approved):

PR #352, opened 2021-07-20
Adding Delegates page with My Delegation data

cool-ant:
Data fetching is largely done. 2 open items of note
 - the way I've been mocking data doesn't allow me to test the logic very well. I'll work that out once we have decent test data.
 - Treasury data isn't included yet; I'll do that in a future PR


brandonfancher:
Not that it has to be a part of this PR, but I do notice that when I have `USE_TEST_DATA` set to `true`, member detail pages break for any member that exists in the fixtures but not in the contract. For example, in the screenshot below, 3 of the members (from May) aren't in the contract, so the app errors out when I click on them.

Again, totally not a big deal right now, but as we're putting polish on this fixtures system (which I really like btw), we'll prob want to address that too.

image

cool-ant:
> Not that it has to be a part of this PR, but I do notice that when I have `USE_TEST_DATA` set to `true`, member detail pages break for any member that exists in the fixtures but not in the contract. For example, in the screenshot below, 3 of the members (from May) aren't in the contract, so the app errors out when I click on them.
> 
> Again, totally not a big deal right now, but as we're putting polish on this fixtures system (which I really like btw), we'll prob want to address that too.
> 
> image

Makes sense. I hadn't ensured all existing stuff continues to work; it's likely live data + fixture data that just doesn't line up. I can make appropriate tweaks for that.
Thanks

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -18,14 +15,39 @@ const getMemberBudgetBalance = () => {
     return {}; // TODO
 };
 
-const getMemberElectionParticipationStatus = () => {
-    return {}; // TODO
-};
+export const getMemberRecordFromName = (
```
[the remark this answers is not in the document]
cool-ant: Since i'm using this in non-UI code, I'm going to leave it as is. I could obviously add a hook that uses the helper function if we think it'll be useful other places. Right now, it's just this same helper function called in 2 parts of the the UI code. Lmk if you feel strongly enough about creating the hook now, and I'll add it. Otherwise, I'll leave it until we find we're needing it more.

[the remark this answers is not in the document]
brandonfancher: Nah, I don't feel strongly about it. That's totally fine. 👍

packages/webapp/src/members/api/members.ts
```diff
@@ -91,7 +91,23 @@ export const getCollectedBy = async (
...
+    bio: "",
+    socialHandles: {},
+    inductionVideo: "",
+    attributions: "",
+    createdAt: 0,
+    representative: "",
+    status: 0,
+    election_participation_status: 0,
+    election_rank: 0,
+};
```
brandonfancher, line 108: Good refactor!

packages/webapp/src/pages/election/index.tsx
```diff
@@ -64,6 +88,22 @@ export const ElectionPage = (props: Props) => {
...
+        members &&
+        loggedInMemberName &&
+        getMemberRecordFromName(members, loggedInMemberName);
+
+    const RSVPStatus = [
+        "no_donation",
+        "in_election",
+        "not_in_election",
+        "recently_inducted",
+    ];
```
brandonfancher, line 102: What would you think about turning this into a [numeric enum](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums)? Then we can reference it in this file, and your fixtures can also use it, instead of using magic numbers?

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
...
+        !leadRepresentative ||
+        !myDelegation
+    ) {
+        return <Text size="lg">Fetching Data...</Text>;
+    }
+
+    const loggedInMember: MemberData | undefined =
+        members && getMemberRecordFromName(members, loggedInMemberName);
+    if (!loggedInMember) {
+        return <Text size="lg">Failed to get loggedInMember record</Text>;
```
brandonfancher, line 76: Might be easier to jump around and test, log out and back in with different users, etc. if you wrap this return and the one above it in a `<RawLayout...` too.

brandonfancher: So I tried this logged in as Egeon and I hit this "Failed to get loggedInMember record" even though he is a member. The reason is that `getMemberRecordFromName` expects a blockchain account name (`egeon.edev`)...not the friendly member name (`Egeon The Great`). `currentMember?.name` is a friendly name.

The only reason it's already working with `Eden Member 11` is that `useCurrentMember` is returning `undefined` because it's actually querying the chain for the associated eden member, but `Eden Member 11` only exists as a fixture, not in the actual chain.

If you get rid of `useCurrentMember` on line `38` and get rid of `loggedInMemberName` on `46` entirely, and instead simply pass in the `accountName` from `37` in lieu of `loggedInMemberName`, it starts working again as expected--for both Eden Member 11 and our friend Egeon.

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
 const MEMBERS_PAGE_SIZE = 18;
```
brandonfancher, line 33: I think we only need fetch 1 here, not 18.

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
...
             <Text size="sm" className="mb-8">
                 Note: Data is in square brackets if it's not JSON (to show if
                 something's undefined)
             </Text>
+            <div>
+                <Text size="lg" className="bg-gray-200">
+                    My Delegation
+                </Text>
+                <Text size="sm" className="mt-4">
+                    {`You [${loggedInMember.account}] are level (rank) ${loggedInMember.election_rank} out of ${membersStats.ranks.length}`}
```
brandonfancher, line 90: When you're in JSX, you don't have to use template strings. You can just do:

`You [{loggedInMember.account}] are level (rank) {loggedInMember.election_rank} out of {membersStats.ranks.length}`

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
...
+                </Text>
+                <Text
+                    size="sm"
+                    className="mt-4"
+                >{`Your Delegation is as follows:`}</Text>
```
brandonfancher, line 95: Same here: `Your Delegation is as follows:`

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
...
+                </Text>
+                <Text size="lg" className="bg-gray-200">
+                    Members
+                </Text>
+                <Text size="sm" className="mb-4">
+                    <code>{`enum for status { pending = 0, active }`}</code>
+                </Text>
+                <Text size="sm" className="mb-4">
+                    <code>
+                        {`enum for election_participation_status { no_donation = 0, in_election, not_in_election, recently_inducted }`}
```
brandonfancher, line 141: and here

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
...
-            <div className="grid grid-cols-2">
+            <div>
+                <Text size="lg" className="mb-4">
+                    -- Raw Table Data --
+                </Text>
+                <Text size="lg" className="bg-gray-200">
+                    Members
+                </Text>
+                <Text size="sm" className="mb-4">
+                    <code>{`enum for status { pending = 0, active }`}</code>
```
brandonfancher, line 137: and here

brandonfancher (changes requested):
I'm not finished reviewing yet...I'll come back to it hopefully tonight or first thing in the morning. But thought I'd go ahead and give you some early feedback. Feel free to go ahead and integrate it.

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,80 +33,128 @@ interface Props {
 const MEMBERS_PAGE_SIZE = 18;
 
 export const DelegatesPage = (props: Props) => {
-    const { isError: isMembersDataFetchError, data: members } = useQuery({
+    const [ualAccount] = useUALAccount();
+    const accountName = ualAccount?.accountName;
+    const { data: currentMember } = useCurrentMember();
+
+    const { data: members } = useQuery({
         ...queryMembers(1, MEMBERS_PAGE_SIZE),
```
brandonfancher, line 41: I believe you can use the `queryMemberData` query here instead (passing in the `accountName`). Then a lot of this becomes quite a bit simpler. You already have the current member's data. You can display their data at the bottom since you're only displaying 1 record there anyway. And if I'm reading this right, you won't even need to invoke your `getMemberRecordFromName` function below either, because you'll already have the `MemberData`.

Also, your queries here don't need the `keepPreviousData` flag since you're not doing any server-side querying. You can get rid of those and the `getServerSideProps` function too.

packages/webapp/src/delegates/api/fixtures.ts
```diff
@@ -4,7 +4,7 @@ export const fixtureMembersStats: MemberStats = {
     active_members: 4,
     pending_members: 0,
     completed_waiting_inductions: 0,
-    ranks: [4, 1],
+    ranks: [1, 4, 1],
 };
 
 export const fixtureMembers: MemberData[] = [
```
brandonfancher, line 10:
```
export const fixtureMembersStats: MemberStats = {
    active_members: 4,
    pending_members: 0,
    completed_waiting_inductions: 0,
    ranks: [1, 4, 1],
};
```

There are 4 active_members. But there are 6 people in the ranks. And then we have 5 active members in the members fixture. Am I reading that right?

Should we consider calculating the values in `fixtureMembersStats` from the values in `fixtureMembers` so that they're always guaranteed to be in sync (assuming we use `fixtureMembers` as the source of truth and derive `fixtureMemberStats` from it?) Or maybe you found that it's better to keep them decoupled and separate so we can more nimbly tweak values? Just want to understand better your thinking on that.

packages/webapp/src/pages/delegates/index.tsx
cool-ant: This is fetching "all" members, so not just one

packages/webapp/src/pages/delegates/index.tsx
cool-ant: I really like using tick quotes everywhere just to keep thing consistent (in case there's a special character like an apostrophe)... :)

packages/webapp/src/delegates/api/fixtures.ts
cool-ant: I've updated the fixtures to be more consistent with notes to describe the dependencies. They were written as I developed each section, so they weren't ever meant to be completely consistent, but I got them all pretty close (with the caveats in comments)

packages/webapp/src/pages/delegates/index.tsx
brandonfancher: You're already not using tick quotes consistently in your JSX. Most strings in your JSX in this file are not wrapped in. `{` ... `}`. 

Besides that...apostrophes work fine in JSX as do most special characters...without the tick quotes. Using string template literals like this goes against React convention and is unnecessarily difficult to write and read.

This is a ReactNode. Basic strings are valid ReactNodes. And the convention is to use strings directly inside JSX. In fact, I'd argue on reason JSX is used in React is to emulate familiar HTML/XML structure. This moves us away from that familiarity.

1. I've never seen anyone do this purposefully in React, unless absolutely necessary.
1. This is one I do actually feel strongly about...hehe.
1. Sparky and I will not be doing this. Your code will be an outlier in this regard.

So I'd urge you to reconsider.

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -18,14 +15,38 @@ const getMemberBudgetBalance = () => {
...
+    loggedInMemberAccount: string | undefined
+): Promise<EdenMember[]> => {
+    let myDelegates: EdenMember[] = [];
 
-export const getMyDelegation = () => {
-    return {}; // TODO
+    if (!loggedInMemberAccount) return myDelegates;
+
+    const leadRepresentative = await getHeadDelegate();
+    let nextDelegate = await getEdenMember(loggedInMemberAccount);
```
brandonfancher, line 34: We should avoid making API network calls directly outside of `react-query`. We lose a lot of benefits (caching, speed, reductions in network traffic, etc.)

You can call these imperatively here using `queryCache.fetchQuery()` or create a custom hook as in this example: https://github.com/eoscommunity/Eden/pull/358.

brandonfancher (changes requested):

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,76 +30,130 @@ interface Props {
...
     });
 
+    const { data: leadRepresentative } = useQuery({
+        ...queryHeadDelegate,
+        keepPreviousData: true,
+    });
+
+    const { data: myDelegation } = useQuery({
+        ...queryMyDelegation(loggedInMember?.account),
+        enabled: !!loggedInMember && !!loggedInMember.account,
```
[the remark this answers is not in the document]
sparkplug0025: yep, easier to read! :D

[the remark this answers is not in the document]
cool-ant: I started in C++... I like symbols. :)
But happy to be more javascripty

packages/webapp/src/pages/election/index.tsx
```diff
@@ -50,6 +65,14 @@ export const ElectionPage = (props: Props) => {
         keepPreviousData: true,
     });
```
brandonfancher, line 66: And I recommend these queries become custom hooks too so we can easily reuse them elsewhere if we want to.

packages/webapp/src/pages/election/index.tsx
```diff
@@ -42,6 +48,15 @@ export const ElectionPage = (props: Props) => {
...
 
+    const { isError: isVoteDataFetchError, data: voteData } = useQuery({
+        ...queryMemberGroupParticipants(
+            loggedInMember?.account,
+            currentElection?.config
+        ),
+        keepPreviousData: true,
+        enabled: !!loggedInMember?.account && !!currentElection?.config,
```
brandonfancher, line 57: `enabled: Boolean(loggedInMember?.account && currentElection?.config),`

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -18,14 +17,57 @@ const getMemberBudgetBalance = () => {
...
-const getMemberElectionParticipationStatus = () => {
-    return {}; // TODO
-};
+const MEMBER_REPRESENTATIVE_IF_NOT_PARTICIPATED_IN_RECENT_ELECTION =
+    "zzzzzzzzzzzzj";
+// check that member has participated in an election (if there's been one yet) (!="zzz...") and came to consensus with their group in the last election (!=0)
+const memberHasRepresentative = (member: EdenMember) =>
+    member.account !== "" &&
+    member.account !==
+        MEMBER_REPRESENTATIVE_IF_NOT_PARTICIPATED_IN_RECENT_ELECTION;
```
brandonfancher, line 26: `const memberHasRepresentative = (member: EdenMember) => member?.account !== MEMBER_REPRESENTATIVE_IF_NOT_PARTICIPATED_IN_RECENT_ELECTION;` ?

brandonfancher (changes requested):
Mostly minor details at this point.

packages/webapp/src/delegates/api/eden-contract.ts
cool-ant: 2 distinct scenarios. I've added another constant to document the difference

packages/webapp/src/delegates/api/fixtures.ts
```diff
@@ -1,15 +1,55 @@
-import { MemberData, MemberStats } from "members";
+import { EdenMember, MemberData, MemberStats } from "members";
```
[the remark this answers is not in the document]
cool-ant: I could go either way on this. We could group the fixtures in one place, but at the same time, they're only used in the sections they're relevant to, so they're pretty nicely encapsulated. The only exception is the members fixture, which I've moved to /members/ to keep the encapsulation uniform. thoughts?

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -133,3 +150,50 @@ export const useIsCommunityActive = () =>
...
+export const useElectionState = () =>
+    useQuery({
+        ...queryElectionState,
+        keepPreviousData: true,
+    });
+
+export const useMemberStats = () =>
+    useQuery({
+        ...queryMembersStats,
+        keepPreviousData: true,
```
brandonfancher, line 198: These 5 new hooks here are using the `keepPreviousData` flag. Why?

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -1,31 +1,92 @@
...
-const getDelegates = () => {
-    return {}; // TODO
+const queryElectionStateHelper = async () => {
+    try {
+        return await queryClient.fetchQuery(
+            queryElectionState.queryKey,
+            queryElectionState.queryFn
+        );
+    } catch (e) {
+        console.error("Error in getMyDelegation:");
```
brandonfancher, line 16: in `getMyDelegation`?

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -22,78 +31,109 @@ interface Props {
...
+                <Text size="lg">Fetching Data...</Text>; [
+                <pre>{JSON.stringify(loggedInMember, null, 2)}</pre>] [
+                <pre>{JSON.stringify(membersStats, null, 2)}</pre>] [
+                <pre>{leadRepresentative}</pre>] [
+                <pre>{JSON.stringify(myDelegation, null, 2)}</pre>] [
+            </RawLayout>
+        );
+    }
+
+    if (!loggedInMember)
```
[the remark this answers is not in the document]
cool-ant: ah, good catch; didn't catch that when the loggedInMember refactor happened

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -1,31 +1,92 @@
...
+            nextDelegate = await queryClient.fetchQuery(queryKey, queryFn);
+            if (!nextDelegate) return myDelegates;
+        }
+        if (
+            nextDelegate.account === leadRepresentative &&
+            memberHasRepresentative(nextDelegate)
+        ) {
+            myDelegates.push(nextDelegate);
+        }
+    } catch (e) {
```
brandonfancher, line 86: I know we talked about this too...but as I think about it more, the worst that happens here if we just let this throw is that we get no data at the query/UI level and an error, right?

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -1,31 +1,92 @@
...
+const queryElectionStateHelper = async () => {
+    try {
+        return await queryClient.fetchQuery(
+            queryElectionState.queryKey,
+            queryElectionState.queryFn
+        );
+    } catch (e) {
+        console.error("Error in getMyDelegation:");
+        console.error(e);
+        return Promise.resolve(undefined);
```
brandonfancher, line 18: I think just returning undefined will resolve the async promise here.

Nonetheless, I'm somewhat uncomfortable with us swallowing all of these errors. React Query expects the async functions it wraps in queries to _throw_. In fact, they even give guidance on how to make async functions that don't normally throw (like `fetch`) to do so: https://react-query.tanstack.com/guides/query-functions#handling-and-throwing-error

I recommend simply removing the `try...catch` here. We're already dealing with the lack of data (`undefined`) in the UI in the event one of these throws. And that just gives us the added option of intercepting and handling errors there in the UI too.

brandonfancher (approved):
Approving pending handling of my remaining/new comments. I suspect @sparkplug0025 would have some valuable feedback too. Might be worth getting his eyes on it.

packages/webapp/src/delegates/api/eden-contract.ts
cool-ant: true

packages/webapp/src/delegates/api/eden-contract.ts
cool-ant: 👍

packages/webapp/src/pages/members/index.tsx
```diff
@@ -53,7 +53,6 @@ export const MembersPage = (props: Props) => {
 
     const { data: memberStats } = useQuery({
         ...queryMembersStats,
-        keepPreviousData: true,
```
cool-ant, line 56: @brandonfancher @sparkplug0025 I believe this `keepPreviousData` isn't needed either, right? This wasn't my addition, which is why I'm asking. I don't want to undo something that was intentional...

sparkplug0025: The `keepPreviousData` helps by caching pagination: https://react-query.tanstack.com/guides/paginated-queries#better-paginated-queries-with-keeppreviousdata

I don't see a reason to remove it... Did I miss something?

packages/webapp/src/pages/_app.tsx
```diff
@@ -27,23 +27,18 @@ dayjs.extend(relativeTime.default);
...
-                <EdenUALProvider>
-                    <Component {...pageProps} />
-                </EdenUALProvider>
-            </Hydrate>
-            <ReactQueryDevtools initialIsOpen={false} />
-            <Toaster />
-        </QueryClientProvider>
-    );
-};
+export const queryClient = new QueryClient();
```
sparkplug0025, line 30: We can't do this. I think I sent you the reason @brandonfancher. From react query docs: 
> Create a new QueryClient instance inside of your app, and on an instance ref (or in React state). This ensures that data is not shared between different users and requests, while still only creating the QueryClient once per component lifecycle.

See more here: https://react-query.tanstack.com/guides/ssr#using-hydration

packages/webapp/src/members/api/eden-contract.ts
```diff
@@ -10,10 +10,15 @@ import {
...
 import { devUseFixtureData } from "config";
-import { fixtureMembersStats } from "delegates/api/fixtures";
+import { fixtureEdenMembers, fixtureMembersStats } from "./fixtures";
 
-export const getEdenMember = (account: string) =>
+export const getEdenMember = (account: string) => {
+    if (devUseFixtureData)
+        return Promise.resolve(
+            fixtureEdenMembers.find((member) => member.account === account)
+        );
```
sparkplug0025, line 19: Cool! I did something very similar in my draft, the difference is that I'm falling back to a regular table call if the fixture is not found: https://github.com/eoscommunity/Eden/pull/361/files#diff-b5c705d489c324900aeac335703bf0ea500c8f433a0e0a328a640c22e2f88303R15-R25

Should we consider that? In any case you can keep like this and then I could merge mine over it if you think that's interesting.

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -133,3 +150,45 @@ export const useIsCommunityActive = () =>
...
+    loggedInMemberAccount: string | undefined
+) => {
+    const { data: currentElection } = useCurrentElection();
+    return useQuery({
+        ...queryMemberGroupParticipants(
+            loggedInMemberAccount,
+            currentElection?.config
+        ),
+        enabled:
+            Boolean(loggedInMemberAccount) && Boolean(currentElection?.config),
```
sparkplug0025, line 182:
```
enabled:
            Boolean(loggedInMemberAccount) && Boolean(currentElection?.config),
```
should be simplified to:
```
enabled: Boolean(loggedInMemberAccount && currentElection?.config)
```

packages/webapp/src/_app/eos/api.ts
```diff
@@ -52,12 +54,22 @@ export const getTableRows = async <T = any>(
...
+    const rows = await getTableRawRows(table, options);
+    // variants are structured as such: array[type: string, <object the variant contains>]
+    // this line is reducing the data to just the data part
+    return rows.map((row) => row[1]);
+};
+
+export const getTableRawRows = async <T = any>(
+    table: string,
+    options?: TableQueryOptions
+): Promise<[string, T][]> => {
```
sparkplug0025, line 66: Nice refactor!

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -1,31 +1,78 @@
...
+
+    while (
+        nextDelegate!.account !== leadRepresentative &&
+        memberHasRepresentative(nextDelegate)
+    ) {
+        myDelegates.push(nextDelegate);
+        const { queryKey, queryFn } = queryMemberByAccountName(
+            nextDelegate!.representative
+        );
+        nextDelegate = await queryClient.fetchQuery(queryKey, queryFn);
```
sparkplug0025, line 68: I think replacing these dangling `queryKey`s & `queryFn`s for a direct query usage like this would be easier:
```
queryClient.fetchQuery(queryMemberByAccountName(nextDelegate!.representative))
```

same for the above `loggedInMemberAccount`

packages/webapp/src/elections/interfaces.ts
```diff
@@ -1,45 +1,75 @@
+import { string } from "zod";
```
sparkplug0025, line 1: Did you really mean to use string from zod? If yes, what is zod behavior in the following interfaces?

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -1,33 +1,163 @@
...
 
-    return rows[0];
+    return rows;
+};
+
+export const getCurrentElection = async () => {
+    if (devUseFixtureData) return fixtureCurrentElection;
+
+    const rawRows = await getTableRawRows<any>(CONTRACT_CURRENT_ELECTION_TABLE);
+    const electionState = rawRows[0][0];
```
sparkplug0025, line 140: What is the data inside the electionState? I suspect its the variant struct name, like `election_st_v0`, and if so why do we need it? Maybe I missed the usage?

packages/webapp/src/pages/members/index.tsx
brandonfancher: Yeah, if it's helping smooth pagination for a pagination-related query, it's relevant. (And I believe pagination has a dependency on this query here.)

Now, if we wanted to, we _could_ switch this out for the new named query--`useMemberStats()`--and keep the `keepPreviousData` flag on that one.

packages/webapp/src/pages/_app.tsx
brandonfancher: You and I discussed it and wondered if the reason for making it a ref was for SSR...but we couldn't find that piece of documentation at the time, so we had concluded we would give this a try. (And I never circled back to run the CURL to see if SSR was working afterwards!)

I'll work with Mike this morning to adjust. I can think of two options:

1. In our named queries dependent on async functions using `queryClient`, do a `useQueryClient` and pass the instance into the async function.
2. Create an API instance within the app component in a ref and give it the `queryClient` instance. The async functions would be methods on that API instance and would therefore have access to the `queryClient`. (That's what I did in a previous production app.)

Option 1 is less intrusive and prob worth option for for now. 

Thoughts, @sparkplug0025?

packages/webapp/src/delegates/api/eden-contract.ts
brandonfancher: `fetchQuery`, unfortunately, doesn't take the object like `useQuery` does.

packages/webapp/src/pages/_app.tsx
cool-ant: Ok, so out of our meeting this morning, I'm not making further changes here until we've researched the issue further.

packages/webapp/src/members/api/eden-contract.ts
cool-ant: @sparkplug0025 This does fall back to a regular table call. That's the structure of everything I've written in this PR. The next line, L20, has the table call.

packages/webapp/src/elections/interfaces.ts
cool-ant: I definitely did *not*. this must have been an autocomplete import I didn't see happen. Great catch @sparkplug0025 !

packages/webapp/src/elections/api/eden-contract.ts
cool-ant: Not sure I understand the question @sparkplug0025 . That *is* the variant struct name. We need this data for election meta info like round start/end time and current election status, eg. seeding, active, final, etc.

packages/webapp/src/members/api/eden-contract.ts
sparkplug0025: Actually it does not, if the fixture flag is on and the fixture data is not present you are returning undefined. (Which is totally fine, I was wondering if the fixture is not present we could still do a table call 😊)

packages/webapp/src/elections/api/eden-contract.ts
sparkplug0025: Yeah, I'm curious to know what you will do with the structure name in the `electionState` variable, I didn't see any usage for it. I'm asking that from the view that all of our other contract table selections we don't do anything with the variant struct name, and then I'm curious to know what you will do with it.

packages/webapp/src/members/api/eden-contract.ts
cool-ant: added in https://github.com/eoscommunity/Eden/pull/370

packages/webapp/src/elections/api/eden-contract.ts
cool-ant: Yeah, the shape of the value that the variant holds is *dependent on* the variant type, and this type (which indicates election status) isn't represented elsewhere, so I've gotta pull it from here so we know election state

PR #353, opened 2021-07-20
Prepare function to publish encrypted messages to the chain

sparkplug0025:
This will be used for publishing secrets to the chain like election meeting links. See more here: https://www.notion.so/edenos/Automating-Video-Conferencing-5e065bd24d1743e689f08d4ed89ea6fc#85310a8b795f44f89e5ad6da3573fb28

This is a sample function with also an example of how the decryption will work.

### DIFF
--- a/packages/webapp/src/_app/eos/secret-publisher.ts
+++ b/packages/webapp/src/_app/eos/secret-publisher.ts
@@ -77,14 +77,14 @@ const ecdhRecipientsKeyEncriptionKeys = async (
     recipientPublicKeys: string[],
     transientPrivateKeyString: string,
     info?: string
-) => {
+): Promise<CryptoKey[]> => {
     const transientPrivateKey = PrivateKey.fromString(
         transientPrivateKeyString
     );
     const transientEc = ec.keyPair({
         priv: new bn(transientPrivateKey.d.toHex(), 16),
     });
-    return await Promise.all(
+    return Promise.all(
         recipientPublicKeys.map(async (publicKey: string) => {
             const recipientEcPublicKey = ec.keyFromPublic(
                 PublicKey.fromString(publicKey).toHex(),
@@ -101,7 +101,10 @@ const ecdhRecipientsKeyEncriptionKeys = async (
     );
 };
 
-const hkdfSha256FromEcdh = async (ecdhSecret: ArrayBuffer, info?: string) => {
+const hkdfSha256FromEcdh = async (
+    ecdhSecret: ArrayBuffer,
+    info?: string
+): Promise<CryptoKey> => {
     const sharedSecretKey = await crypto.subtle.importKey(
         "raw",
         ecdhSecret,
@@ -157,7 +160,7 @@ const encryptSessionKeys = async (sessionKey: CryptoKey, keks: CryptoKey[]) => {
 const encryptMessage = async (
     sessionKey: CryptoKey,
     message: string
-): Uint8Array => {
+): Promise<Uint8Array> => {
     const encodedMessage = new TextEncoder().encode(message);
     const encryptedMessage = await crypto.subtle.encrypt(
         {
--- a/packages/webapp/src/_app/eos/secret-publisher.ts
+++ b/packages/webapp/src/_app/eos/secret-publisher.ts
@@ -1,0 +1,217 @@
+/**
+ * recipient_ecc_pubkey = eden.keys_table( where account = logged_in user )
+check if recipient_ecc_pubkey matches browser
+
+transient_ecc_pubkey = new pubkey (same curve as recipient_ecc_pubkey)
+
+recipient_key1 = eden.keys_table( where account = recipient1 user )
+recipient_key2 = eden.keys_table( where account = recipient2 user )
+recipient_key3 = eden.keys_table( where account = recipient3 user )
+
+// each recipient
+key_encryption_key = HKDF-SHA256(ECDH(recipient_pubkey, transient_privkey))
+
+session_key = random AES-128 key
+
+// each recipient
+encrypted_session_key = AES-KW(KEK_recipient, session key)
+
+encrypted_link = AES-GCM(session_key, ZOOM_LINK)
+
+eden.publishelection(
+  publisher = logged_in_user
+  round_no = '2',
+  encrypted_link,
+// for each recipient
+  [(encrypted_session_keyN, recipient_keyN, transient_ecc_pubkey)]
+)
+ */
+
+// import { generateKeyPair, PrivateKey, PublicKey } from "eosjs";
+// // import {generateKeyPair, PrivateKey, PublicKey, sha256, Signature} from '../eosjs-key-conversions';
+// import { KeyType } from "eosjs/dist/eosjs-numeric";
+import ecc, { PrivateKey, PublicKey } from "eosjs-ecc";
+const ec = require("elliptic").ec("secp256k1");
+const bn = require("bn.js");
+
+export const publishSecretToChain = async (
+    message: string,
+    publisherAccount: string,
+    recipientAccounts: string[]
+) => {
+    const [publisherKey, ...recipientKeys] = await fetchRecipientKeys([
+        publisherAccount,
+        ...recipientAccounts,
+    ]);
+    console.info(publisherKey, recipientKeys);
+
+    // TODO: check publisherkey matches the current one in browser
+
+    const transientKeyPair = await generateRandomTransientKey();
+    console.info(transientKeyPair);
+
+    const keks = await ecdhRecipientsKeyEncriptionKeys(
+        [publisherKey, ...recipientKeys],
+        transientKeyPair.privateKey
+    );
+    console.info(keks);
+
+    const sessionKey = await generateRandomSessionKey();
+    console.info(sessionKey);
+
+    const encryptedSessionKeys = await encryptSessionKeys(sessionKey, keks);
+    console.info(encryptedSessionKeys);
+    const encryptedMessage = await encryptMessage(sessionKey, message);
+    console.info(encryptedMessage);
+
+    const obj = {
+        encryptedSessionKeys,
+        recipientKeys,
+        transientKeyPair,
+        encryptedMessage,
+    };
+    console.info(obj);
+    return obj;
+};
+
+const fetchRecipientKeys = async (accounts: string[]) => {
+    // TODO: real implementation
+    // fetch eden.keys_table( where account = each account item )
+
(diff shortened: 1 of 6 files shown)


packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,217 @@
...
+    );
+    const derivedKey = await crypto.subtle.deriveBits(
+        {
+            name: "HKDF",
+            hash: "SHA-256",
+            salt: new Uint8Array([]),
+            info: new Uint8Array([]),
+        },
+        sharedSecretKey,
+        256
```
swatanabe, line 140: This will end up generating an AES-256 key.  It's not a security problem, but it is a bit wasteful as we only have 128-bit security through the whole system.  Also, I think it would be simpler to use deriveKey and avoid having to importKey.

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,217 @@
...
+        ["encrypt", "decrypt"]
+    );
+
+    return await Promise.all(
+        keks.map(async (rawKek: Uint8Array) => {
+            console.info("importing kek");
+            const kek = await crypto.subtle.importKey(
+                "raw",
+                rawKek,
+                { name: "HKDF" },
```
swatanabe, line 175: Is this right?  kek should be an AES-128 key and have wrapKey as the keyUsage.

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,217 @@
...
+        sessionKey,
+        { name: "AES-GCM" },
+        false,
+        ["encrypt"]
+    );
+
+    const encryptedMessage = await crypto.subtle.encrypt(
+        {
+            name: "AES-GCM",
+            iv: sessionKey, // TODO: is it secure for our use cases?
```
swatanabe, line 210: I don't trust this construct.  A zero initialization vector would be better.

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,217 @@
...
+        { name: "HKDF" },
+        false,
+        ["deriveKey", "deriveBits"]
+    );
+    const derivedKey = await crypto.subtle.deriveBits(
+        {
+            name: "HKDF",
+            hash: "SHA-256",
+            salt: new Uint8Array([]),
+            info: new Uint8Array([]),
```
swatanabe, line 137: The info field can be set to something unique for each meeting.

packages/webapp/src/_app/eos/secret-publisher.ts
sparkplug0025: Yeah, I think in general I can just stop exporting and re-importing keys, I was going to do that in my final refactor! thanks for the 128-bit heads up. Will change it!

packages/webapp/src/_app/eos/secret-publisher.ts
sparkplug0025: fixed!

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,171 @@
...
+    });
+    return await Promise.all(
+        recipientPublicKeys.map(async (publicKey: string) => {
+            const recipientEcPublicKey = ec.keyFromPublic(
+                PublicKey.fromString(publicKey).toHex(),
+                "hex"
+            );
+            const ecdhSecret = transientEc.derive(
+                recipientEcPublicKey.getPublic()
+            );
```
sparkplug0025, line 95: @tbfleming I think it's a little nasty here since I'm using eosjs-ecc to handle eos keys but then converting it to `elliptic. secp256k1` format to derive the ecdh secret. Is there a more appropriated approach for doing this?

tbfleming: eosjs-ecc is deprecated. eosjs replaced it with elliptic in 21.0.2. See the release notes for eosjs's conversion helpers: https://github.com/EOSIO/eosjs/releases/tag/v21.0.2

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,174 @@
...
+    const transientPrivateKey = PrivateKey.fromString(
+        transientPrivateKeyString
+    );
+    const transientEc = ec.keyPair({
+        priv: new bn(transientPrivateKey.d.toHex(), 16),
+    });
+    return Promise.all(
+        recipientPublicKeys.map(async (publicKey: string) => {
+            const recipientEcPublicKey = ec.keyFromPublic(
+                PublicKey.fromString(publicKey).toHex(),
```
swatanabe, line 90: What will happen if publicKey uses the P-256 curve instead of secp256k1?

sparkplug0025: nice point. Do we need to enforce that all recipients have K1? what if recipients have different types of public keys? I think for now we should drive the type through the UI and generate the encryption keys as K1!

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,239 @@
...
+    transientPublicKey: string,
+    encryptedSessionKey: Uint8Array,
+    info?: string
+) => {
+    const recipientPrivateKey = retrieveRecipientPrivateKey(recipientPublicKey);
+    const ecdhSecret = deriveEcdhSecret(
+        recipientPrivateKey,
+        PublicKey.fromString(transientPublicKey)
+    );
+    console.info("ecdh secret is ", ecdhSecret);
```
swatanabe, line 79: This is fine for debugging, but in production, cryptographic secrets should never be logged.

sparkplug0025: I plan to remove all the console prints

packages/webapp/src/_app/eos/secret-publisher.ts
tbfleming: `package.json`, `yarn.lock`, and `eosjs-ecc.d.ts` still pull in eosjs-ecc

sparkplug0025: Nice catch! Removed!

packages/webapp/src/_app/eos/secret-publisher.ts
tbfleming: The new code depends on all key types being the same:

```
const deriveEcdhSecret = (
    eosPrivateKeyA: PrivateKey,
    eosPublicKeyB: PublicKey
) => {
    return eosPrivateKeyA
        .toElliptic()
        .derive(eosPublicKeyB.toElliptic().getPublic())
        .toArrayLike(ArrayBuffer); // TODO: review endianness, empty = Big Endian
};
```

This will work if both keys are `k1` or both are `r1`.

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,240 @@
...
+};
+
+const deriveEcdhSecret = (
+    eosPrivateKeyA: PrivateKey,
+    eosPublicKeyB: PublicKey
+) => {
+    return eosPrivateKeyA
+        .toElliptic()
+        .derive(eosPublicKeyB.toElliptic().getPublic())
+        .toArrayLike(ArrayBuffer); // TODO: review endianness, empty = Big Endian
```
tbfleming, line 136: Since this is treated as pseudo-random data by HKDF, the endianness doesn't matter

swatanabe: If we want to allow third-party front ends to interact with this data (or if we later decide to re-write this using a different API), then it matters, because every implementation needs to use the same endianness.

sparkplug0025:
It seems like we have two concerns:
1. Key curves need to be the same (K1 or R1). If any participant is using a different interface for Eden they need to make sure they are also using the same curve as we are. (K1).
2. Another third-party interface could be deriving ECDH secrets. They need to follow the same endianess as we are: Big Endian.

I will add both points in this file doc and also to our Automating Video Conferencing page. Anything else to be reviewed and approved?

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,252 @@
+import { PrivateKey, PublicKey } from "eosjs/dist/eosjs-jssig";
+import { generateKeyPair } from "eosjs/dist/eosjs-key-conversions";
+import { KeyType } from "eosjs/dist/eosjs-numeric";
+
+/**
+ * Provides a way to encrypt data to be published on chain. Useful for
+ * secretly sharing any temporal data. Eg. induction meeting links, election
```
swatanabe, line 7: I think you mean temporary data.

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,252 @@
...
+ * room links.
+ *
+ * IMPORTANT NOTES:
+ * - Any data on chain will eventually be decrypted. Never use it for PII (read the
+ * above description for reasanoble use cases).
```
swatanabe, line 12: reasonable.

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -0,0 +1,252 @@
...
+import { generateKeyPair } from "eosjs/dist/eosjs-key-conversions";
+import { KeyType } from "eosjs/dist/eosjs-numeric";
+
+/**
+ * Provides a way to encrypt data to be published on chain. Useful for
+ * secretly sharing any temporal data. Eg. induction meeting links, election
+ * room links.
+ *
+ * IMPORTANT NOTES:
+ * - Any data on chain will eventually be decrypted. Never use it for PII (read the
```
swatanabe, line 11: If encrypted data were not decrypted, there would be no point to storing it at all.  With that being said, the level of security is not suitable for long-term protection of high-value secrets.

swatanabe (approved):
LGTM, but I'm not a JS programmer.

packages/webapp/src/_app/eos/secret-publisher.ts
sparkplug0025: Yeah here I meant more of decrypted by badactors/attackers. I will rephrase and add your point.

tbfleming (approved):
LGTM

PR #355, opened 2021-07-21
Basic UI for My Delegation

brandonfancher:
Available for now at the `/delegates/ui` path. This is all faked data...not even fixtures yet.

This set the stage for:
* Hooking up to delegation APIs and testing more extensively with fixture data
* Adding blank states (no election has happened yet, logged out, etc.)
* Move to the main `/delegates` path (or wherever its final resting place will be)

But those can be done in subsequent PRs.

### Screenshot
image


packages/webapp/src/members/components/temp-member-card-resign-modal.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from "react";
```
cool-ant, line 1: Filename is "temp-"? Is that intentional?

packages/webapp/src/pages/delegates/ui/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+                            />
+                        </Container>
+                    )}
+                </div>
+            ))}
+        </>
+    );
+};
+
+const levels: string[] = [
```
cool-ant, line 70: We should either have a TODO comment here or make this dynamically determined.

packages/webapp/src/pages/election/components.tsx
```diff
@@ -57,13 +57,17 @@ export const MembersPage = (props: Props) => {
...
                 {members.isLoading && "Loading members..."}
                 {members.error && "Fail to load members"}
             </Container>
             <MembersGrid members={members.data?.slice(1, 2)}>
                 {(member) => (
-                    <WinningMemberChip key={member.account} member={member} />
+                    <DelegateChip
+                        key={member.account}
+                        member={member}
+                        level="D3 - Chief Delegate"
```
cool-ant, line 69: Similar here w the hard coded value

packages/webapp/src/_app/ui/delegate-badge.tsx
```diff
@@ -0,0 +1,12 @@
+import { FaStar } from "react-icons/fa";
+
+export const DelegateBadge = ({ size }: { size: number }) => (
```
sparkplug0025, line 3: Should we keep our size pattern? eg for buttons:
`export type ButtonSize = "sm" | "md" | "lg";`

packages/webapp/src/members/components/temp-member-card-resign-modal.tsx
sparkplug0025: Probably also rename component from `MemberCard` to `MemberCardResignModal`

sparkplug0025: and update/remove the data-testid

packages/webapp/src/pages/delegates/ui/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+            dehydratedState: dehydrate(queryClient),
+        },
+    };
+};
+
+interface Props {
+    delegatesPage: number;
+}
+
+const MEMBERS_PAGE_SIZE = 4;
```
sparkplug0025, line 23: I don't think this is relevant for delegate. If it's related to the logged user we will not have the logged user session in our server, so hydrating it here does not make sense. (Unless we plan to implement user sessions handling in the server)

sparkplug0025: tldr; this block can be removed and we can let the client execute the query as it's doing below

packages/webapp/src/members/components/temp-member-card-resign-modal.tsx
brandonfancher: Yeah, this wasn't meant to me checked in. Sorry, guys!

packages/webapp/src/pages/delegates/ui/index.tsx
brandonfancher: I'll add a TODO. We'll make it dynamic when we hook up to the data.

packages/webapp/src/_app/ui/delegate-badge.tsx
brandonfancher: I don't know...yet. For right now, I see this more like a composed react-icon where size is flexibly variable. But if we come up with a strong convention on badge use and sizing, then yes, I think we make this more opinionated.

packages/webapp/src/pages/delegates/ui/index.tsx
brandonfancher: I'll remove the boilerplate for now. I erred on the side of leaving stuff in for us to remove when we hook this UI up vs. making everything perfectly clean. We are going to need to decide what we want on this page when a member is not logged in. If, for example, we want to show the Head Chief / Satoshi on this page, that could potentially be SSR'd.

packages/webapp/src/members/components/temp-member-card-resign-modal.tsx
brandonfancher: Removed.

packages/webapp/src/pages/election/components.tsx
brandonfancher: Yes. This is just example UI for now. Not hooked up to any data. We'll make sure these are dynamic and are a matter of configuration. 👍

packages/webapp/src/pages/delegates/ui/index.tsx
brandonfancher: Done

cool-ant (approved):

PR #358, opened 2021-07-23
Dirty example of a useMyDelegation hook.

brandonfancher:
This is just an example. Not intending to merge.

### DIFF
--- a/packages/webapp/src/delegates/hooks.ts
+++ b/packages/webapp/src/delegates/hooks.ts
@@ -1,0 +1,51 @@
+import { useEffect, useState } from "react";
+import { useQuery, useQueryClient } from "react-query";
+
+import {
+    queryHeadDelegate,
+    queryMemberByAccountName,
+    useCurrentMember,
+} from "_app";
+import { EdenMember } from "members";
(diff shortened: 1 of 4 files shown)


packages/webapp/src/delegates/hooks.ts
```diff
@@ -0,0 +1,51 @@
...
+        (async () => {
+            while (
+                nextDelegate.account !== leadRepresentative &&
+                memberHasRepresentative(nextDelegate)
+            ) {
+                myDelegates.push(nextDelegate);
+                const { queryKey, queryFn } = queryMemberByAccountName(
+                    nextDelegate!.representative
+                );
+                nextDelegate = await queryClient.fetchQuery(queryKey, queryFn); // TODO: try...catch
```
brandonfancher, line 33: We can't use hooks (`useQuery`) in loops, and I don't think `useQueries` will work well with this kind of `while` loop. So instead, I query imperatively. (This should be wrapped in a `try...catch`.) Because it's async, and we can't use async...await directly in a hook, we wrap it in an IIFE and use it to build up state (`delegates`) which we ultimately return to the calling component or hook.

PR #359, opened 2021-07-26
Brandon/election UI enhancements

brandonfancher:
### What's in this PR?
* Updates styles to match latest designs.
* Adds leaderboard list ordering animation example. 

### The Animation
The idea is that the frontend will poll (easy with `react-query`) periodically to get the latest vote totals for each participant in the round. When those totals update in the UI (or when you submit a new vote), the order of participants on the leaderboard will change along with the vote totals. You'll also see who is voting for whom.

This POC demonstrates that statically when you (the logged-in user) vote, as long as you're logged in as one of the people in the list.

https://user-images.githubusercontent.com/7911424/127066493-c0a5a763-c950-4b97-85f9-fb856d77051a.mp4

### DIFF
--- a/packages/webapp/src/_app/ui/expander.tsx
+++ b/packages/webapp/src/_app/ui/expander.tsx
@@ -5,27 +5,30 @@ import { Container } from "_app";
 
 interface ExpanderProps {
     header: React.ReactNode;
-    hideContentDivider?: boolean;
+    showContentDivider?: boolean;
     startExpanded?: boolean;
+    inactive?: boolean;
     children: React.ReactNode;
 }
 
 export const Expander = ({
     header,
-    hideContentDivider,
+    showContentDivider,
     startExpanded = false,
+    inactive,
     children,
 }: ExpanderProps) => {
     const [isExpanded, setIsExpanded] = useState<boolean>(startExpanded);
     const containerClass =
         "flex justify-between items-center group cursor-pointer hover:bg-gray-100 active:bg-gray-200 select-none";
     const contentDividerClass =
-        !isExpanded || hideContentDivider ? "" : "border-b border-gray-100";
+        isExpanded && showContentDivider ? "border-b border-gray-100" : "";
     return (
-        <div>
+        <div className={inactive ? "bg-gray-50" : ""}>
             <Container
                 className={`${containerClass} ${contentDividerClass}`}
                 onClick={() => setIsExpanded(!isExpanded)}
+                darkBg={inactive}
             >
                 {header}
                 {isExpanded ? (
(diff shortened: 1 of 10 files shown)


packages/webapp/src/_app/ui/expander.tsx
```diff
@@ -5,27 +5,30 @@ import { Container } from "_app";
 
 interface ExpanderProps {
     header: React.ReactNode;
-    hideContentDivider?: boolean;
+    showContentDivider?: boolean;
```
cool-ant, line 8: 👍

packages/webapp/src/_app/ui/expander.tsx
```diff
@@ -5,27 +5,30 @@ import { Container } from "_app";
...
     startExpanded?: boolean;
+    inactive?: boolean;
     children: React.ReactNode;
 }
 
 export const Expander = ({
     header,
-    hideContentDivider,
+    showContentDivider,
```
cool-ant, line 16: Does we use a particular form of this enough to provide a default value?

packages/webapp/src/elections/components/election-member-chips.tsx
```diff
@@ -9,12 +9,17 @@ interface VotingMemberChipProps {
...
+    votingFor?: string;
 }
 
 export const VotingMemberChip = ({
     member,
     onSelect,
     isSelected,
+    votesReceived,
+    votingFor,
+    ...rest
```
cool-ant, line 22: same here

packages/webapp/src/pages/election/ui.tsx
```diff
@@ -223,16 +223,32 @@ const OngoingRoundSegment = ({
...
-                        onSelect={() => onSelectMember(member)}
-                    />
-                )}
-            </MembersGrid>
+            <Flipper flipKey={sortMembersByVotes}>
+                <MembersGrid members={sortMembersByVotes}>
+                    {(member) => (
+                        <Flipped
+                            key={`leaderboard-${member.account}`}
+                            flipId={`leaderboard-${member.account}`}
```
cool-ant, line 231: really clean. Love how this reads

packages/webapp/src/_app/ui/expander.tsx
brandonfancher: It defaults to `undefined` right now, which is falsy. BUT...I think adding a default value in does make it more explicit...easier to see what's actually going on. Added.

packages/webapp/src/_app/ui/generic-member-chip.tsx
```diff
@@ -19,27 +19,30 @@ export const GenericMemberChip = ({
     actionComponent,
     contentComponent,
     footerComponent,
+    ...rest
```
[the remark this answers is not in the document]
brandonfancher: Good suggestion. Updating to `...containerProps`.

PR #361, opened 2021-07-27
Basics of Setting & ReEntering the Public Encryption Password

sparkplug0025:
- If the election password was never set:
image

image

- If the election password is set but not present:
image

image


packages/webapp/src/pages/oauth/zoom.tsx
```diff
@@ -75,16 +76,18 @@ export default ZoomOauthPage;
...
         setZoomAccountJWT(undefined);
     };
 
     const generateZoomMeetingLink = async () => {
-        const x = await publishSecretToChain(
+        const x = await encryptSecretForPublishing(
             "https://us05web.zoom.us/j/81089675368?pwd=MUFWVkowNmk1SFVLTkJ4eEtyY01ldz09",
             ualAccount.accountName,
             ["participanta", "participantb", "participantc"],
+            queryClient,
```
brandonfancher, line 90: Opting for the dependency-injection route here (vs. simply importing the queryClient?)

packages/webapp/src/encryption/components/encryption-password-alert.tsx
```diff
@@ -0,0 +1,369 @@
...
+        updateEncryptionPassword,
+    } = useEncryptionPassword();
+
+    if (!encryptionPassword) {
+        return null;
+    }
+
+    const promptNewKey =
+        promptSetupEncryptionKey && !encryptionPassword.publicKey;
+    if (promptNewKey) {
```
brandonfancher, line 38: I'm seeing this prompt even when signed out completely.

brandonfancher: Also, I see this for a quick moment while the page is loading (if I refresh). When this turns orange or red or whatever color it is, that flash will be quite noticeable.

So this should only display:

1. When a user is signed in,
2. when the user is an active eden member
3. we have their eden member record and see that they don't have a key set

brandonfancher (approved):
Just a few minor comments. Otherwise, works great! I set a key on `edenmember11`, so that one's burned now. Let me know and I'll send you the private key.

I imagine password manager integration (like getting LastPass to prompt for saving for this side) will come later? I saw some basic integration (Lastpass icon in the field, so it recognizes it) but not sure why it doesn't prompt me to save once I submit.

packages/webapp/src/pages/oauth/zoom.tsx
sparkplug0025: nice catch! when I did that we didn't have the global one! (changed now!)

PR #362, opened 2021-07-27
More UI Enhancements for Ongoing Elections

brandonfancher:
## In this PR

- [x] Add all participants to previous election rounds
- [x] Add examples of election video upload play buttons
- [x] Add round timer UI

## UI Goals

image

## Later (in a different PR)
* Consens-o-meter


packages/webapp/src/_app/hooks/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+    useEffect(() => {
+        const tick = () => {
+            savedCallback.current();
+        };
+        if (delay !== null) {
+            const id = setInterval(tick, delay);
+            return () => clearInterval(id);
+        }
+    }, [delay]);
+};
```
cool-ant, line 21: Really nice.

brandonfancher: Thanks! (I stole it from the internet.)

packages/webapp/src/_app/ui/pie-chart.tsx
```diff
@@ -0,0 +1,57 @@
+interface Props {
+    percent: number;
+    size: number;
+    backgroundColor?: string;
+    fillColor?: string;
+}
+
+export const PieChart = ({
```
cool-ant, line 8: I'm being a bit picky, but this isn't a pie *chart* (or at least isn't much of one). I'd like to name it a bit more accurately. Either call it a pie-status-indicator or pie-countdown-clock. or a countdown pie-mer. lol. no idea, but a name that wouldn't conjure pictures of multiple differently-colored segments

packages/webapp/src/elections/components/countdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <PieChart percent={percent} size={28} />
+            <div className="flex-1 ml-2">
+                <Text size="sm" className="font-semibold">
+                    Vote Timer
+                </Text>
+                <Text size="sm">{timeRemainingString}</Text>
+            </div>
+        </div>
+    );
+};
```
cool-ant, line 46: I'm liking Pie-mer. :)

packages/webapp/src/elections/components/election-member-chips.tsx
```diff
@@ -72,7 +83,23 @@ interface DelegateChipProps {
...
+    member: MemberData;
+    delegateLevel?: string;
+    isDelegate?: boolean;
+    electionVideoCid?: string;
+}
+
+export const ElectionParticipantChip = ({
+    member,
+    delegateLevel,
+    isDelegate = false,
```
cool-ant, line 100: isDelegate is minimally used. Is there a future use for it? Or could you just default delegateLevel = 0 and use that as the flag in the code?

cool-ant (approved):

packages/webapp/src/elections/components/countdown.tsx
brandonfancher: ✅

packages/webapp/src/_app/ui/pie-chart.tsx
brandonfancher: Yeah, I hear you. This component is not the countdown and has nothing to do with time, but it's not a full-fledged pie chart either.

packages/webapp/src/elections/components/election-member-chips.tsx
brandonfancher: Yeah, in fact, if we look at `GenericMemberChip` and `ElectionParticipantChip`, I hint with TODOs on that prop is likely to change or go away depending on the shape of the data coming in and whether we can simply rely on that. This was an easy, in-your-face, way to demonstrate the application of the delegate badge...but we'll likely get that from the member object when we hook it up.

PR #363, opened 2021-07-28
Add election video upload

swatanabe:


contracts/eden/src/members.cpp
```diff
@@ -219,6 +219,13 @@ namespace eden
       });
    }
 
+   bool members::can_upload_video(uint8_t round, eosio::name member)
+   {
+      auto iter = member_tb.find(member.value);
+      return iter != member_tb.end() && iter->election_rank() >= round &&
+             iter->representative() != eosio::name(-1);
```
tbfleming, line 226: If the last round that a user was in has been tallied (user is no longer in `vote_tb`), it's still within the 48-hour window, but the member was in a group that didn't reach consensus, it looks like that member can't upload a video for either that round or any previous round.

swatanabe: In that case, what I think should happen is:
- representative == eosio::name(0)
- election_rank == highest round reached,
which should allow this to pass

tbfleming (approved):

PR #378, opened 2021-08-04
intRA-round member grouping

cool-ant:
This PR adds to election/data.tsx to show the basics of getting member grouping **for** and **during** the round that's currently happening. A follow-on PR will handle the inter-election / previous-round scenarios.

### DIFF
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -28,6 +28,8 @@ import {
     getCurrentElection,
     getElectionState,
     getMemberGroupParticipants,
+    getVoteData,
+    getVoteDataRow,
 } from "elections/api/eden-contract";
 import { ActiveStateConfigType } from "elections/interfaces";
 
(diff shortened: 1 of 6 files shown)


packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -93,15 +95,16 @@ export const getMemberGroupParticipants = async (
     return rows;
 };
 
-const getVoteDataRow = async (
-    opts: VoteDataQueryOptionsByField
+export const getVoteDataRow = async (
+    opts?: VoteDataQueryOptionsByField
```
brandonfancher, line 99: You could make `opts` required here, as this function really _does_ need it.

Then in your `queryVoteDataRow` in `queries.ts`, you can throw there: 

```
export const queryVoteDataRow = (account?: string) => ({
    queryKey: ["query_vote_data_row", account],
    queryFn: () => {
        if (!account) throw new Error("getVoteDataRow requires an account (not undefined)"); // or something like this
        return getVoteDataRow({ fieldName: "member", fieldValue: account });
    },
});
```

Whether you handle `undefined` for `account` in `queryVoteDataRow` or later in `getVoteDataRow`, I recommend you throw and error and not just return `undefined`. React Query expects it. When we use the query in our UI, we'll still get `undefined` for data, but now we'll have a friendly error to go along with it. (And the query is disabled anyways if `account` is `undefined`, so we should never see it.

This would tighten up the types and make it super clear to the developer what's happening instead of just seeing `undefined`.

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -54,8 +55,16 @@ export const queryMemberGroupParticipants = (
 ) => ({
     queryKey: ["query_member_group_participants", memberAccount, config],
     queryFn: () => getMemberGroupParticipants(memberAccount, config),
+    enabled: Boolean(memberAccount),
 });
 
+export const queryVoteDataRow = (account: string | undefined) => ({
```
brandonfancher, line 61: Optional: use the undefined shorthand `(account?: string)` for these. But they do the same, as you know.

packages/webapp/src/elections/api/eden-contract.ts
brandonfancher: This could apply to `getMemberGroupParticipants` above too.

brandonfancher (approved):
Looks good. A couple minor comments.

packages/webapp/src/_app/hooks/queries.ts
cool-ant: and another great one. Thanks for pointing this out. This makes a lot of sense and tightens up plenty of the code... Thanks!

PR #380, opened 2021-08-05
Adds ParticipationCard for Election Registration

sparkplug0025:
Adds the Election participation flow to participate and cancel participation. My next iteration will be to detect if the password is present or not and merge the set encryption password before confirming the election.

image
image
image
image

### DIFF
--- a/packages/webapp/src/elections/api/fixtures.ts
+++ b/packages/webapp/src/elections/api/fixtures.ts
@@ -30,6 +30,12 @@ export const fixtureElectionState: ElectionState = {
     last_election_time: "2021-01-16T16:00:00.000",
 };
 
+export const fixtureRegistrationElection: CurrentElection = {
+    electionState: "registration?", // TODO: confirm?
+    start_time: "2022-01-16T16:00:00.000Z",
+    election_threshold: 100,
+};
+
 // This data reflects an in-progress election round and, therefore,
 // won't be consistent with other fixture data
 // In the grand scheme of other fixture data,
--- a/packages/webapp/src/elections/api/fixtures.ts
+++ b/packages/webapp/src/elections/api/fixtures.ts
@@ -31,8 +31,8 @@ export const fixtureElectionState: ElectionState = {
 };
 
 export const fixtureRegistrationElection: CurrentElection = {
-    electionState: "registration?", // TODO: confirm?
-    start_time: "2022-01-16T16:00:00.000Z",
+    electionState: "current_election_state_registration",
+    start_time: "2022-01-16T16:00:00.000",
     election_threshold: 100,
 };
 
--- a/packages/webapp/src/elections/components/participation-card.tsx
+++ b/packages/webapp/src/elections/components/participation-card.tsx
@@ -1,0 +1,287 @@
+import { useState } from "react";
+import dayjs from "dayjs";
+
+import {
+    Button,
+    Container,
+    ElectionParticipationStatus,
+    Form,
+    Heading,
+    Modal,
+    onError,
+    queryMemberByAccountName,
+    Text,
+    useCurrentElection,
+    useCurrentMember,
+    useUALAccount,
+} from "_app";
+// import { generateEncryptionKey } from "_app/eos/secret-publisher";
+// import { setEncryptionPublicKeyTransaction } from "encryption/transactions";
+
+// import { UpdateEncryptionPassword, useEncryptionPassword } from "../hooks";
+import { useEffect } from "react";
+import { getElectionStartDateTime } from "elections/utils";
+import { setElectionParticipation } from "elections/transactions";
+import { useQueryClient } from "react-query";
+
+interface Props {
+    promptSetupEncryptionKey?: boolean;
+}
+
+export const ParticipationCard = ({ promptSetupEncryptionKey }: Props) => {
+    const [ualAccount] = useUALAccount();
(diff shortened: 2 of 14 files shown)


packages/webapp/src/elections/api/fixtures.ts
```diff
@@ -31,6 +31,12 @@ export const fixtureElectionState: ElectionState = {
     last_election_time: "2022-01-16T16:00:00.000",
 };
 
+export const fixtureRegistrationElection: CurrentElection = {
+    electionState: "registration?", // TODO: confirm?
+    start_time: "2022-01-16T16:00:00.000Z",
```
brandonfancher, line 36: So, the data from the API comes back _without_ the `Z` at the end. That means that when we're using dev fixtures, the UI will account for time zone properly (because you have the `Z` here) but when we fetch from the API, the UI will not properly adjust for time zone, as there's no `Z`.

I recommend we handle this by appending the `Z` manually (and in the same place whether fixtures or fetching). Maybe in `getCurrentElection`?

I'm open to suggestions.

packages/webapp/src/elections/components/participation-card.tsx
```diff
@@ -0,0 +1,287 @@
...
+    Text,
+    useCurrentElection,
+    useCurrentMember,
+    useUALAccount,
+} from "_app";
+// import { generateEncryptionKey } from "_app/eos/secret-publisher";
+// import { setEncryptionPublicKeyTransaction } from "encryption/transactions";
+
+// import { UpdateEncryptionPassword, useEncryptionPassword } from "../hooks";
+import { useEffect } from "react";
```
brandonfancher, line 22: It appears `useEffect` is not being used.

packages/webapp/src/elections/components/participation-card.tsx
```diff
@@ -0,0 +1,287 @@
...
+import { getElectionStartDateTime } from "elections/utils";
+import { setElectionParticipation } from "elections/transactions";
+import { useQueryClient } from "react-query";
+
+interface Props {
+    promptSetupEncryptionKey?: boolean;
+}
+
+export const ParticipationCard = ({ promptSetupEncryptionKey }: Props) => {
+    const [ualAccount] = useUALAccount();
```
brandonfancher, line 32: It appears `ualAccount` is not being used.

packages/webapp/src/elections/components/participation-card.tsx
```diff
@@ -0,0 +1,287 @@
...
+            );
+
+            close();
+        } catch (error) {
+            console.error(error);
+            onError(error);
+        }
+
+        setIsLoading(false);
+    };
```
brandonfancher, line 254: Almost all of this `onSubmit` is like the one above it. Could be shared...although I realize this is a work in progress and you may be consciously duplicating them for now.

packages/webapp/src/elections/interfaces.ts
```diff
@@ -61,7 +61,7 @@ export type CurrentElection = CurrentElectionBase &
         | CurrentElection_activeState
         | CurrentElection_initVotersState
         | CurrentElection_postRoundState
-        | CurrentElection_registationState
+        | CurrentElection_registrationState
```
brandonfancher, line 64: good catch!

brandonfancher (approved):
A couple comments to address, but approving so you're not blocked.

brandonfancher:
One more note; we should make sure that when determining whether to prompt the user for their password or not, we also check the _validity_ of their password/private key in the browser against the public key on chain...make sure they match. If they do not, the user should be asked to re-enter their password. (If we're already doing this check, I don't see it.) 

This goes for both the check in the registration flow _and_ the banners that appear elsewhere.

packages/webapp/src/elections/components/participation-card.tsx
```diff
@@ -0,0 +1,287 @@
...
+        return <Text>Error parsing the Election start date.</Text>;
+    }
+
+    const parsedElectionStartDateTime = dayjs(electionStartDateTime);
+    const parsedParticipationTimeLimit = parsedElectionStartDateTime.subtract(
+        24,
+        "hour"
+    );
+    const parsedEstimatedElectionEndDateTime = parsedElectionStartDateTime.add(
+        10, // TODO: is it dynamic?
```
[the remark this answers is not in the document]
sparkplug0025: I will leave this fixed value with this TODO, so we can adjust and calculate it dynamically based on how many RSVP confirmations we have since I'm not sure how to get it yet.

packages/webapp/src/elections/api/fixtures.ts
sparkplug0025: Nice Catch! Yeah I will replicate the fixture without Z and concatenate it manually!

packages/webapp/src/elections/components/participation-card.tsx
sparkplug0025: Exactly that, because the other onsubmit will need to consider password confirmation!

packages/webapp/src/elections/components/participation-card.tsx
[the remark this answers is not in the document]
cool-ant: which could be estimated and minimized by querying community size / RSVPs, calculating the number of rounds needed (rounding up by one if we're near a cross-over point), then planning an hour for each round

PR #381, opened 2021-08-05
Index members by representative

swatanabe:


contracts/eden/include/members.hpp
```diff
@@ -68,11 +72,15 @@ namespace eden
...
-      EDEN_FORWARD_FUNCTIONS(value, primary_key)
+      EDEN_FORWARD_FUNCTIONS(value, primary_key, by_representative)
    };
    EOSIO_REFLECT(member, value)
 
-   using member_table_type = eosio::multi_index<"member"_n, member>;
+   using member_table_type = eosio::multi_index<
+       "member"_n,
+       member,
+       eosio::indexed_by<"byrep"_n,
```
tbfleming, line 82: The table exists in production: https://www.bloks.io/account/genesis.eden?loadContract=true&tab=Tables&table=member&account=genesis.eden&scope=0&limit=100

swatanabe: Right... For some reason, I was thinking that it was okay because the index uses fields from member_v1 which doesn't exist in production yet.

tbfleming: It'd be a win if we add a function to multi_index that rescans a table (at a provided scope value) and populates a new index

swatanabe: I'm working on that now.

swatanabe: Looking at the code, it looks as though multi_index will mostly work as long as you don't try to use the new index.  The only case that fails is a modify that changes the key used for the new secondary index.

contracts/eden/include/members.hpp
```diff
@@ -37,6 +37,10 @@ namespace eden
       std::optional<eosio::public_key> encryption_key;
 
       uint64_t primary_key() const { return account.value; }
+      uint128_t by_representative() const
```
cool-ant, line 40: Since the election_rank is built into this index, perhaps we should call this `by_group` or something more accurate. It's not really by_representative, right?

libraries/eosiolib/contracts/include/eosio/multi_index.hpp
```diff
@@ -2314,5 +2326,34 @@ namespace eosio
...
+                typename index_type::secondary_key_type>::db_idx_find_primary(_code.value, _scope,
+                                                                              index_type::name(),
+                                                                              iter->primary_key(),
+                                                                              secondary);
+            if (existing < 0)
+            {
+               secondary_index_db_functions<typename index_type::secondary_key_type>::db_idx_store(
+                   _scope, index_type::name(), payer.value, iter->primary_key(), secondary);
+            }
+            next_primary_key = iter->primary_key() + 1;
```
tbfleming, line 2354: wraps if a pk exists that's 0xffff...

libraries/eosiolib/contracts/include/eosio/multi_index.hpp
```diff
@@ -2314,5 +2326,34 @@ namespace eosio
...
+      {
+         using namespace _multi_index_detail;
+
+         using index_type = decltype(get_index<IndexName>());
+         for (auto iter = lower_bound(next_primary_key), end = this->end();
+              max_steps > 0 && iter != end; --max_steps, ++iter)
+         {
+            auto secondary = index_type::extract_secondary_key(*iter);
+            auto existing = secondary_index_db_functions<
+                typename index_type::secondary_key_type>::db_idx_find_primary(_code.value, _scope,
```
tbfleming, line 2345: For some types, overwrites `secondary` with garbage when not found

libraries/eosiolib/contracts/include/eosio/multi_index.hpp
swatanabe: That isn't possible in the current usage, since the pk of the member table must be an account that exists.

contracts/eden/include/members.hpp
swatanabe: I don't see why it isn't by representative.  The election_rank is just there to disambiguate.

tbfleming (approved):

PR #383, opened 2021-08-05
Restyle password prompt banner; eliminate flicker.

brandonfancher:
## In this PR
1. The password prompt would flicker on load/refresh--even if the password was present for the signed-in user in the browser. I eliminated this by adding an `isInitialized` state flag to the hook. (But I'm interested, @sparkplug0025, if you see a better way of achieving this--I tried several.)
2. This moves the election password prompt to the layout as a full-width banner and restyles it.

image


sparkplug0025 (approved):
Another approach could be adding a flag on the Encryption Key Interface but I think the isInitialized works perfectly well being able to merge with isLoading from query is pretty nice! 👌

PR #387, opened 2021-08-06
Style up the election screen when election is pending

brandonfancher:
## What's in this PR?
* Styles up the election opt-in/out page
* Adds signed out state
* Adds non-member state
* Adds FAQ placeholder from xd designs (still need to fill those in with content)

### Screenshots

image

image

image


cool-ant (approved):
Really clean. Nice

PR #388, opened 2021-08-06
Post-round: Get member group

cool-ant:


packages/webapp/src/members/api/eden-contract.ts
```diff
@@ -13,10 +13,13 @@ import { devUseFixtureData } from "config";
 import { fixtureEdenMembers, fixtureMembersStats } from "./fixtures";
 
 export const getEdenMember = (account: string) => {
+    console.info(`getEdenMember().account[${account}]`);
     if (devUseFixtureData) {
         const edenMember = fixtureEdenMembers.find(
             (member) => member.account === account
         );
+        console.info(`devFixtureData; edenMember:`);
+        console.info(edenMember);
```
cool-ant, line 22: leaving these in until I add the no-consensus scenario; then I'll remove them

packages/webapp/src/members/api/fixtures.ts
```diff
@@ -446,3 +446,19 @@ export const fixtureMembersStats: MemberStats = {
...
     ranks: [7, 4, 1, 1, 1],
 };
+
+export const fixtureMembersInGroup = (
+    round: number,
+    delegateAccount: string
+) => {
+    console.info(
+        `fixtureMembersInGroup().round[${round}], delegateAccount[${delegateAccount}]`
+    );
```
cool-ant, line 456: same here

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -145,6 +152,51 @@ const getVoteDataRows = async (
 
 export const getVoteData = getVoteDataRows;
 
+const getCommonDelegateAccountForGroupWithThisMember = (
+    round: number,
+    member?: EdenMember
```
brandonfancher, line 157: Looks like everything calling this function _has_ the `member`. Functions calling this function have the member as a required param. Does this need to be optional here?

brandonfancher (approved):

packages/webapp/src/elections/api/eden-contract.ts
cool-ant: good call; question mark removed

PR #389, opened 2021-08-09
Add Encryption password detection to Participation

sparkplug0025:
- Extracted the password forms from the encryption password alert
- Reused them in the participation when the password are required to confirm
- If the user has the encryption key in the browser the participation will NOT prompt for entering the key again

### DIFF
--- a/packages/webapp/src/elections/components/registration-election-components/participation-card.tsx
+++ b/packages/webapp/src/elections/components/registration-election-components/participation-card.tsx
@@ -210,27 +210,6 @@ const ConfirmParticipationModal = ({ isOpen, close }: ModalProps) => {
         await onSubmit(setEncryptionPasswordTrx);
     };
 
-    const renderStep = () => {
-        switch (step) {
-            case ParticipationStep.ConfirmParticipation:
-                return (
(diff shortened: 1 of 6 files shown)


packages/webapp/src/elections/components/registration-election-components/participation-card.tsx
```diff
@@ -161,6 +196,41 @@ const ConfirmParticipationModal = ({ isOpen, close }: ModalProps) => {
...
+                return (
+                    <ConfirmPasswordStep
+                        onSubmit={submitPasswordConfirmation}
+                        isLoading={isLoading}
+                        onCancel={close}
+                    />
+                );
+        }
+    };
+
```
cool-ant, line 233: Unless we expect more steps than this in the future, I think this is small enough and nearly-only-components enough that we could move this rendering code right into the <Modal /> rendering below with just `step === ParticipationStep.ConfirmParticipation && ... <ConfirmParticipationStep ... />` and `step === ...ConfirmPassword ... <ConfirmPasswordStep ... />`

optional, but I think it reads better together

cool-ant (approved):

PR #390, opened 2021-08-09
Adding the No Consensus group scenario

cool-ant:


packages/webapp/src/pages/election/data.tsx
```diff
@@ -212,15 +212,23 @@ export const ElectionPage = () => {
                 {/* TODO: remove the true below once done */}
                 {(true || !Boolean(voteRowForLoggedInMember)) && (
```
brandonfancher, line 213: Are you still working on this or should the `true` go away?

brandonfancher (approved):
Looks good! May have more questions as we implement and test against the live test election data, but we'll cross that bridge if/when those questions arise.

packages/webapp/src/pages/election/data.tsx
cool-ant: debatable at this point, especially since the assumption is test data, but we can always mod fixtures locally to actuate this. So... removed. thx

PR #391, opened 2021-08-09
Ongoing Election Hookup - Real API!

brandonfancher:
Purpose of this PR is to hook up _existing_ ongoing-election UI to real, life contract table data. 

## In This PR
- [x] Get in-election round group queries to work
- [x] Tweak ongoing-election UI to match data coming in better
- [x] Vote action
- [x] Hook up previous round data (if round came to consensus)
- [x] Polling for votes / UI magically updates

## Future Work
- How do we display (during an election) prior rounds that did _not_ come to consensus? Can we query that?
- Differentiate Chief Delegates round (label as Chief Delegates, no voting, explainer, etc.)
- Signed out / non-member view
- Views for someone who doesn't become a delegate at a various points during the election
- Bigger refactors
- etc.

In this PR, we're prioritizing functionality. In other words, I just want to get it working. We will refactor later. So be gentle. ;)

### DIFF
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -79,7 +79,12 @@ export const queryParticipantsInCompletedRound = (
     electionRound?: number,
     member?: EdenMember
 ) => ({
-    queryKey: ["query_current_election", electionRound, member],
+    queryKey: [
+        "query_current_election",
+        electionRound,
+        member,
+        isStillParticipating,
+    ],
     queryFn: () => {
         if (typeof electionRound !== "number" || !member)
             throw new Error(
@@ -225,10 +230,8 @@ export const useHeadDelegate = () =>
         ...queryHeadDelegate,
     });
 
-export const useParticipantsInCompletedRound = (
-    electionRound?: number,
-    member?: EdenMember
-) => {
+export const useParticipantsInCompletedRound = (electionRound?: number) => {
+    const { data: member } = useCurrentMember();
     const { data: voteData } = useVoteDataRow(member?.account);
     console.info("voteData:", voteData);
 
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -258,6 +258,7 @@ export const useMemberGroupParticipants = (memberAccount?: string) => {
         ...queryMemberGroupParticipants(memberAccount, currentElection?.config),
         enabled: Boolean(memberAccount && currentElection?.config),
         refetchInterval: 10000,
+        refetchIntervalInBackground: true,
     });
 };
 
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -76,26 +76,17 @@ export const queryVoteData = (options: VoteDataQueryOptionsByGroup = {}) => ({
 });
 
 export const queryParticipantsInCompletedRound = (
-    isStillParticipating: boolean,
-    electionRound?: number,
-    member?: EdenMember
+    electionRound: number,
+    member?: EdenMember,
+    voteData?: VoteData
 ) => ({
-    queryKey: [
-        "query_current_election",
-        electionRound,
-        member,
-        isStillParticipating,
-    ],
+    queryKey: ["query_current_election", member, voteData, electionRound],
     queryFn: () => {
-        if (typeof electionRound !== "number" || !member)
+        if (!member)
             throw new Error(
-                "useParticipantsInCompletedRound() requires a value for 'memberAccount' and 'electionRound'"
+                "useParticipantsInCompletedRound() requires a value for 'memberAccount'"
             );
-        return getParticipantsInCompletedRound(
-            electionRound,
-            member,
-            isStillParticipating
-        );
+        return getParticipantsInCompletedRound(electionRound, member, voteData);
     },
 });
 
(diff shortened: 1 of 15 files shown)


packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -213,13 +230,20 @@ export const useHeadDelegate = () =>
...
-    electionRound?: number,
-    member?: EdenMember
-) => {
+export const useParticipantsInCompletedRound = (electionRound?: number) => {
+    const { data: member } = useCurrentMember();
+    const { data: voteData } = useVoteDataRow(member?.account);
+    console.info("voteData:", voteData);
+
+    const isStillParticipating = Boolean(voteData);
+    console.info(`isStillParticipating[${isStillParticipating}]`);
```
brandonfancher, line 239: We use `isStillParticipating` as a way to see whether you're the winner of the previous round or not, but what if the election is ongoing and in Round 3? You won Round 1 and lost Round 2. And you want to look at Round 1 results. You're waiting for the election to end and observing. We would be unable to show you that you won Round 1, wouldn't we? (Maybe we have to compare Round 1 with Round 2 to find the common participant to show you that you're the winner of Round 1?)

@cool-ant

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -181,12 +199,16 @@ export const getParticipantsInCompletedRound = async (
...
-        limit: 6,
-    });
+    const addWinner = isStillParticipating ? [member] : [];
+
+    return addWinner.concat(
+        await getTableRows(CONTRACT_MEMBER_TABLE, {
+            index_position: 2,
+            key_type: "i128",
+            lowerBound,
+            limit: 6,
```
brandonfancher, line 209: We're ending up with 7 results in our returned array no matter what the group size here. (6 limit plus winner). Options:

* Find a way to get the `upperBound` here.
* Have a sufficiently high limit and filter out those whose rep is not the winner. (but this would only work if there _is_ a winner, I believe).

packages/webapp/src/_app/hooks/queries.ts
brandonfancher: Also, if you're not signed in as the winner, the winner is getting omitted from the list entirely.

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -138,55 +146,79 @@ const getVoteDataRows = async (
...
+    let commonDelegate = member.representative;
+    if (electionRank > roundRequested) {
+        commonDelegate = member.account;
+    } else if (electionRank < roundRequested) {
+        throw new Error(
+            "Cannot fetch round participants in round member did not participate in."
+        );
+    }
+
     return isValidDelegate(commonDelegate) ? commonDelegate : "";
```
cool-ant, line 177: If this is working its way into a useQuery, we should probably throw rather than return "", right? Empty delegate account doesn't have meaning to the rest of the app

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -138,55 +146,79 @@ const getVoteDataRows = async (
...
+    member: EdenMember,
+    voteData?: VoteData
+): Promise<{ participants: EdenMember[]; delegate?: string } | undefined> => {
     const commonDelegate = getCommonDelegateAccountForGroupWithThisMember(
         electionRound,
-        member
+        member,
+        voteData
     );
+    if (!commonDelegate) return undefined;
```
cool-ant, line 190: Then that above change may affect this line

packages/webapp/src/elections/api/fixtures.ts
```diff
@@ -31,7 +36,7 @@ export const fixtureElectionState: ElectionState = {
 };
 
 export const fixtureRegistrationElection: CurrentElection = {
-    electionState: "current_election_state_registration",
+    electionState: ElectionStatus.Registration,
```
cool-ant, line 39: nice catch

packages/webapp/src/elections/components/ongoing-election-components/completed-round-segment.tsx
```diff
@@ -1,35 +1,37 @@
...
+} from "_app";
 import { Button, Container, Expander } from "_app/ui";
 import { ElectionParticipantChip } from "elections";
 import { MembersGrid } from "members";
 
 import RoundHeader from "./round-header";
 
 interface CompletedRoundSegmentProps {
-    round: number;
+    roundIndex: number;
```
cool-ant, line 14: *really* nice rename

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
```diff
@@ -83,30 +75,49 @@ export const OngoingRoundSegment = ({
...
+            const transaction = setVote(
+                authorizerAccount,
+                roundIndex,
+                selectedMember?.account
+            );
+            console.info("signing trx", transaction);
+            const signedTrx = await ualAccount.signTransaction(transaction, {
+                broadcast: true,
+            });
+            console.info("electopt trx", signedTrx);
```
cool-ant, line 100: remove these console.logs?

cool-ant (approved):

packages/webapp/src/elections/api/eden-contract.ts
brandonfancher: Agreed. Good catch!

packages/webapp/src/elections/api/eden-contract.ts
brandonfancher: In fact, removes it altogether!

PR #394, opened 2021-08-10
Completes fn to generate meeting and publish to Chain

sparkplug0025:
- Completes the link between the encryption function and the meeting generation in the zoom oauth page.
- The function generateZoomMeetingLink can be reused from our correct elections flow (as opposed to the current dummy example)
- Then right after we should use `encryptSecretForPublishing` and `setInductMeeting` to publish to the chain.

PS: Should work after #393 abi is deployed. We got an abi parameters issue.

### DIFF
--- a/packages/webapp/src/_app/eos/secret-publisher.ts
+++ b/packages/webapp/src/_app/eos/secret-publisher.ts
@@ -46,7 +46,7 @@ export const encryptSecretForPublishing = async (
     const transientKeyPair = generateEncryptionKey();
     console.info(transientKeyPair);
 
-    const keks = await ecdhRecipientsKeyEncriptionKeys(
+    const keks = await ecdhRecipientsKeyEncryptionKeys(
         publicKeys,
         transientKeyPair.privateKey,
         info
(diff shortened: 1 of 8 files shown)


packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -47,11 +47,11 @@ export const encryptSecretForPublishing = async (
     console.info(transientKeyPair);
 
     const keks = await ecdhRecipientsKeyEncriptionKeys(
```
cool-ant, line 49: spelling; "EncryptionKeys"

packages/webapp/src/_app/eos/secret-publisher.ts
```diff
@@ -61,25 +61,40 @@ export const encryptSecretForPublishing = async (
...
         publisherKey,
         transientKeyPair.publicKey.toString(),
         encryptedSessionKeys[0],
         info
     );
+    if (decryptedMessage !== message) {
+        throw new Error(
+            "an error occurred during encryption/decryption of the data"
+        );
+    }
```
cool-ant, line 84: Decided to not remove this test? Or is this an oversight?

cool-ant (approved):

packages/webapp/src/_app/eos/secret-publisher.ts
sparkplug0025: No, I actually wanted to reinforce we are able to decrypt the message before returning it. The only thing to do is remove the logs for sensitive content.

PR #395, opened 2021-08-10
Make Our code Great Again

cool-ant:
This one's just to make me happier. These have been bothering, so I took the few minutes it was going to take to clean this stuff up and make it prettier.

This PR may be a mess until we merge Brandon's open PR, on which I'll rebase

### DIFF
--- a/packages/webapp/src/_app/eos/interfaces.ts
+++ b/packages/webapp/src/_app/eos/interfaces.ts
@@ -1,8 +1,19 @@
+export type TableKeyType =
+    | "name"
+    | "i64"
+    | "i128"
+    | "i256"
+    | "float64"
+    | "float128"
+    | "sha256"
+    | "ripemd160";
+
+type TableIndexBound = number | string | null;
 export interface TableQueryOptions {
     scope?: string;
-    lowerBound?: any;
-    upperBound?: any;
+    lowerBound?: TableIndexBound;
+    upperBound?: TableIndexBound;
     limit?: number;
     index_position?: number;
-    key_type?: string;
+    key_type?: TableKeyType;
 }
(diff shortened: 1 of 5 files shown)


packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -133,14 +130,14 @@ export const getVoteDataRow = async (
...
+    opts: TableQueryOptions
 ): Promise<VoteData[] | undefined> => {
     if (devUseFixtureData)
         return Promise.resolve(
             fixtureVoteDataRows.filter(
                 (vote) =>
-                    vote.index >= opts.lowerBound &&
-                    vote.index <= opts.upperBound
+                    vote.index >= opts.lowerBound! &&
+                    vote.index <= opts.upperBound!
```
cool-ant, line 140: I would probably do something different here if this weren't purely test data / dev time, but it will never affect production

brandonfancher (approved):

packages/webapp/src/_app/eos/interfaces.ts
```diff
@@ -1,8 +1,19 @@
+export type TableKeyType =
+    | "name"
+    | "i64"
+    | "i128"
+    | "i256"
+    | "float64"
+    | "float128"
+    | "sha256"
+    | "ripemd160";
```
sparkplug0025, line 9: nice!

sparkplug0025 (approved):
super great refactor! especially by killing all the any types! 

Another code that will be hugely simplified is https://github.com/eoscommunity/Eden/blob/sparkplug0025/zoom-request/packages/webapp/src/inductions/api/eden-contract.ts#L49-L93

I see you are using 1 index per table name but there are tables with multiple indexes like the induction one. Maybe it's worth thinking about a table and index position as the key of your `TABLE_INDEXES` map. 

Maybe not since it only affect inductions and probably we don't want to impact them now. Unless you have bandwith to, then I'd love to see both induction and elections tackled.

Else, this is great! Amazing job here! :D

PR #401, opened 2021-08-12
Pre-Election Transitions & UI Enhancements

brandonfancher:
## In This PR
- Handle pre-election state transitions, including automating transition of UI into the `Active` state of the election.
- Adjust UI and add niceties, such as countdown timer.
- Factor our `useCountdown` hook. Callback from that hook triggers polling for updates to election state. That polling stops when election enters `Active` state.
- Prevent accidental dismissal of registration modal when on the password creation/confirmation screen.
- Add timezone codes to times.
- Add delay after opting in/out so that blockchain has time to catch up before updating UI.
- Style up copy button on password form.

### DIFF
--- a/packages/webapp/src/_app/hooks/utils.ts
+++ b/packages/webapp/src/_app/hooks/utils.ts
@@ -21,7 +21,7 @@ export const useInterval = (callback: () => void, delay: number | null) => {
 };
 
 const now = () => new Date();
-const padTime = (time: number) => (time < 10 ? `0${time}` : time);
+const padTime = (time: number) => time.toString().padStart(2, "0");
 
 interface Countdown {
     startTime: Date;
--- a/packages/webapp/src/_app/hooks/utils.ts
+++ b/packages/webapp/src/_app/hooks/utils.ts
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from "react";
+import { useEffect, useState, useRef } from "react";
 
 export const useInterval = (callback: () => void, delay: number | null) => {
     const savedCallback = useRef<() => void>(() => {});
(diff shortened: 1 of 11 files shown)


packages/webapp/src/_app/hooks/utils.ts
```diff
@@ -19,3 +19,41 @@ export const useInterval = (callback: () => void, delay: number | null) => {
...
+const now = () => new Date();
+const padTime = (time: number) => (time < 10 ? `0${time}` : time);
+
+interface Countdown {
+    startTime: Date;
+    endTime: Date;
+    onEnd?: () => void;
+}
+
+export const useCountdown = ({ startTime, endTime, onEnd }: Countdown) => {
```
brandonfancher, line 32: Most of the code in this hook was cut and pasted from the PieMer component. Changes to hone in on when reviewing: `hrRemaining` addition and return value for `hmmss`.

packages/webapp/src/_app/hooks/utils.ts
```diff
@@ -19,3 +19,41 @@ export const useInterval = (callback: () => void, delay: number | null) => {
         }
     }, [delay]);
 };
+
+const now = () => new Date();
+const padTime = (time: number) => (time < 10 ? `0${time}` : time);
```
cool-ant, line 24: Alternative: time.padStart(2, "0")

cool-ant (approved):

packages/webapp/src/_app/hooks/utils.ts
brandonfancher: Nice! Thank you!

PR #404, opened 2021-08-12
Ongoing Election Transitions & States

brandonfancher:
This makes our current election UI more functional. When timers end or election states change, the UI reacts on its own. Loading and error states are handled better. Language is tweaked.

It's the end of the day on Friday. I would have liked to have had more time to review my work here. But I think it's review-worthy now.

The usual disclaimers: This incorporates several minor refactors. Obviously, we'll want a more in-depth refactor after we deliver this milestone to our users. But we're getting there!

### In This PR
- [x] Improve ongoing election error and loading states.
- [x] Display real names for "Voting for..." text on voting participant chips.
- [x] Factor out `<VotingRoundParticipants />` component (starting to break this monster apart).
- [x] Create `useTimeout()` hook in anticipation of timer bonanza.
- [x] Config option for zoom call duration.
- [x] Automatic transitions: round warmup -> meeting -> round finalization -> new round.
- [x] Appropriate timers for each of these stages.
- [x] UI handles those transitions in a more user-friendly way.
- [x] Surface voting errors in toast.

### Later PRs
- One more pass to update to design tweaks in XD
- Make Meeting Call/Zoom button in election UI work (@sparkplug0025 is tackling that in #405)
- No consensus view (@cool-ant is working on that)
- Redirect user to My Delegation page when election completes
- More refactors


packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -1,32 +1,46 @@
...
+        return <Ongoing.ErrorLoadingElection />;
     }
 
+    const { election_round_time_sec, election_break_time_sec } = globals;
+    const roundDurationSec = election_round_time_sec + election_break_time_sec;
+    const roundDurationMs = roundDurationSec * 1000;
+    const roundIndex = election.round ?? memberStats.ranks.length;
+    const roundEndTimeRaw = election.round_end ?? election.seed.end_time;
+    const roundEndTime = dayjs(roundEndTimeRaw + "Z");
+    const roundStartTime = dayjs(roundEndTime).subtract(roundDurationMs);
```
brandonfancher, line 42: This is a good candidate for factoring out.

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
```diff
@@ -1,79 +1,160 @@
...
+
+    const [stage, setStage] = useState<RoundStage>(currentStage);
+
+    const isVotingOpen = [RoundStage.Meeting, RoundStage.PostMeeting].includes(
+        stage
+    );
+
+    useTimeout(() => {
+        setStage(stage + 1);
+    }, timeRemainingToNextStageMs);
```
brandonfancher, line 88: A lot of this is also a good candidate for factoring out. I could see a hook...or at least a util coming from this.

packages/webapp/src/elections/components/ongoing-election-components/round-header.tsx
```diff
@@ -1,73 +1,35 @@
...
-          headlineText?: string;
-      };
+type RoundHeaderProps = {
+    isRoundActive: boolean;
+    headlineComponent: React.ReactNode;
+    sublineComponent: React.ReactNode;
+    children?: React.ReactNode;
+};
 
 export const RoundHeader = ({
```
brandonfancher, line 13: This poor component was just trying to do _too_ much. Made him more presentational.

packages/webapp/src/_app/hooks/utils.ts
```diff
@@ -23,13 +23,38 @@ export const useInterval = (callback: () => void, delay: number | null) => {
 const now = () => new Date();
 const padTime = (time: number) => time.toString().padStart(2, "0");
 
+export const useTimeout = (callback: () => void, delay: number | null) => {
```
sparkplug0025, line 26: nit: null -> undefined

packages/webapp/src/config.ts
```diff
@@ -115,6 +115,10 @@ export const validUploadActions: ValidUploadActions = {
     },
 };
 
+export const electionMeetingDurationMs =
+    Number(process.env.NEXT_PUBLIC_ELECTION_MEETING_DURATION_MS) ||
```
sparkplug0025, line 119: please add `NEXT_PUBLIC_ELECTION_MEETING_DURATION_MS` in the above log and variable check, which will make you to add to .env file too. (It will help us on deploying/debugging in another environments)

packages/webapp/src/elections/components/ongoing-election-components/completed-round-segment.tsx
```diff
@@ -29,16 +29,7 @@ export const CompletedRoundSegment = ({
...
-                <RoundHeader
-                    roundIndex={roundIndex}
-                    subText={
-                        winner
-                            ? `Delegate elect: ${winner.name}`
-                            : "Consensus not achieved"
-                    }
-                />
-            }
+            header={<Header roundIndex={roundIndex} winner={winner?.name} />}
```
sparkplug0025, line 32: +1

packages/webapp/src/elections/components/ongoing-election-components/round-header.tsx
sparkplug0025: +1!

packages/webapp/src/elections/components/ongoing-election.tsx
sparkplug0025: yeah probably create another get dates utility function in the utils.ts returning all of these values in an object. (can be a later pr)

packages/webapp/src/pages/election/index.tsx
```diff
@@ -20,38 +20,36 @@ export const ElectionPage = () => {
...
         case ElectionStatus.Final: // TODO: This is one state where there's a board but no satoshi. UI should reflect that.
-            return <OngoingElection />;
+            return <OngoingElection election={election} />;
         default:
-            return (
-                <Container>
-                    <Heading size={2}>Processing election</Heading>
-                </Container>
-            );
+            return <LoaderSection />;
```
sparkplug0025, line 47: do we have a way to get no election for some query error and keep showing the loader forever? should we account errors? (not very critical I think)

sparkplug0025 (approved):

packages/webapp/src/_app/hooks/utils.ts
brandonfancher: The `delay` is designed to explicitly be set to `null`, which clears the timeout. I'll add a comment linking to where I got this from (https://usehooks-typescript.com/react-hook/use-timeout) but it follows the same interface as Dan Abramov's `setInterval`, which we also have in our codebase and which also takes a `number | null` for this.

packages/webapp/src/config.ts
brandonfancher: In the interest of time, I'll go ahead and do this. But are you opposed to the idea of something like this in the logs?

`ELECTION_MEETING_DURATION_MS="${process.env.NEXT_PUBLIC_AA_FETCH_AFTER || "[NOT SET - DEFAULTING TO 40 MIN]"}"`

That way, we don't have to keep adding to the already sizable number of env vars?

brandonfancher: (And good catch, btw. Def needs to be logged out.)

packages/webapp/src/pages/election/index.tsx
brandonfancher: Yeah, I added fetch error handling elsewhere, but not here at this top level. Good catch. In this case, I think what would happen is that the loading spinner _and_ the contents would just go away and the user would be left wondering what happened. I'll add a small error message here.

PR #419, opened 2021-08-18
Rebased ongoing election changes

cool-ant:


packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -1,12 +1,16 @@
...
+import { BsInfoCircle } from "react-icons/bs";
 
 import {
+    useUALAccount,
     useCommunityGlobals,
     useCurrentMember,
     useMemberStats,
-    useUALAccount,
+    useMyDelegation,
+    useOngoingElectionData as useOngoingElectionData,
```
brandonfancher, line 10: drop the redundant `as useOngoingElectionData`

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -354,3 +413,39 @@ export const useEncryptedData = (scope: EncryptionScope, id: string) =>
...
+    );
+    return useQuery<Election, Error>({
+        queryKey,
+        queryFn,
+        enabled:
+            Boolean(loggedInMember) &&
+            Boolean(memberStats) &&
+            Boolean(electionState) &&
+            Boolean(myDelegation),
+    });
```
brandonfancher, line 450: I think we could shorten this to `Boolean(loggedInMember && memberStats && electionState && myDelegation),`.

Do we need the check for `loggedInMember` and should we add a check for the presence of `votingMemberData` instead?

packages/webapp/src/_app/api/interfaces.ts
```diff
@@ -9,3 +11,30 @@ export enum MemberStatus {
...
+    areRoundsWithNoParticipation: boolean;
+    // .length === number of rounds that have completed (regardless of current member's participation)
+    completedRounds: ElectionCompletedRound[];
+    ongoingRound: {
+        participants: EdenMember[];
+        participantsMemberData: MemberData[];
+        projectedDelegate?: string;
+        isSortitionRound?: boolean;
+    };
+}
```
brandonfancher, line 40: We should consider moving these to `src/elections/interfaces`. I think the only reason `ElectionParticipationStatus` ended up here is that it's applied to members and not exclusively elections. Maybe there's a case to be made for moving it too, though.

packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -27,9 +32,13 @@ export const OngoingElection = ({ election }: { election: any }) => {
         isLoading: isLoadingMemberStats,
         isError: isErrorMemberStats,
     } = useMemberStats();
+    const { data: loggedInVoteData } = useVoteDataRow(loggedInUser?.account);
+    const { data: myDelegationSoFar } = useMyDelegation();
+    const { data: ongoingRoundData } = useOngoingElectionData({
+        enabled: Boolean(loggedInUser) && Boolean(memberStats),
```
brandonfancher, line 38: we already spoke about several of the unused variables and calls in this file.

brandonfancher (approved):

PR #420, opened 2021-08-18
The election of no consensus

cool-ant:
adding the various other states to Election screen: no participation by member, no delegate for member, completed rounds, ongoing round

This is all for functionality and testing; cleanup will happen in follow-on PR

### DIFF
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -42,7 +42,6 @@ import {
     CurrentElection_activeState,
     VoteData,
 } from "elections/interfaces";
-import { EncryptionScope, getEncryptedData } from "encryption";
 
 export const queryHeadDelegate = {
     queryKey: "query_head_delegate",
@@ -331,7 +330,6 @@ export const useMemberGroupParticipants = (
             currentActiveElection?.config
         ),
         ...queryOptions,
-        enabled,
     });
 };
 
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -41,6 +41,7 @@ import {
     CurrentElection,
     CurrentElection_activeState,
     Election,
+    ElectionStatus,
     VoteData,
 } from "elections/interfaces";
 import { EncryptionScope, getEncryptedData } from "encryption/api";
@@ -312,18 +313,17 @@ export const useMemberGroupParticipants = (
     roundIndex?: number,
     queryOptions: any = {}
 ) => {
-    // console.info(`useMGP().top roundIndex[${roundIndex}]`);
+    console.info(`   --->    useMGP().top roundIndex[${roundIndex}]`);
     const { data: currentElection } = useCurrentElection();
     // ASSUMPTION: this use method will only be called by *non*-Chief ongoing rounds
     const currentActiveElection = currentElection as CurrentElection_activeState;
 
-    let enabled =
-        Boolean(memberAccount && currentActiveElection?.config) &&
-        roundIndex !== undefined;
+    let enabled = Boolean(memberAccount) && roundIndex !== undefined;
 
     if ("enabled" in queryOptions) {
         enabled = enabled && queryOptions.enabled;
     }
+    console.info(`useMGP().enabled[${enabled}]`);
 
     return useQuery<VoteData[], Error>({
         ...queryMemberGroupParticipants(
@@ -430,6 +430,12 @@ export const useOngoingElectionData = (): UseQueryResult<
     );
     let { data: votingMemberData } = useMemberDataFromVoteData(
         membersInOngoingRound
+    );
+    console.info(
+        "membersInOngoingRound:",
+        membersInOngoingRound,
+        "votingMemberData:",
+        votingMemberData
     );
 
     const { queryKey, queryFn } = queryOngoingElectionData(
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -43,7 +43,7 @@ import {
     Election,
     VoteData,
 } from "elections/interfaces";
-import { EncryptionScope } from "encryption/api";
+import { EncryptionScope, getEncryptedData } from "encryption/api";
 
 export const queryHeadDelegate = {
     queryKey: "query_head_delegate",
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -12,7 +12,7 @@ import {
     VoteDataQueryOptionsByGroup,
     MemberStats,
 } from "members";
-import { Election, getCommunityGlobals } from "_app/api";
+import { getCommunityGlobals } from "_app/api";
 
 import { useUALAccount } from "../eos";
 import {
@@ -40,8 +40,10 @@ import {
     ActiveStateConfigType,
     CurrentElection,
     CurrentElection_activeState,
+    Election,
     VoteData,
 } from "elections/interfaces";
+import { EncryptionScope } from "encryption/api";
 
 export const queryHeadDelegate = {
     queryKey: "query_head_delegate",
@@ -330,6 +332,7 @@ export const useMemberGroupParticipants = (
             currentActiveElection?.config
         ),
         ...queryOptions,
+        enabled,
     });
 };
 
@@ -411,9 +414,9 @@ export const useEncryptedData = (scope: EncryptionScope, id: string) =>
         ...queryEncryptedData(scope, id),
         enabled: Boolean(id),
     });
-export const useOngoingElectionData = (
-    queryOptions: any = {}
-): UseQueryResult<Election | undefined> => {
+export const useOngoingElectionData = (): UseQueryResult<
+    Election | undefined
+> => {
     const { data: loggedInMember } = useCurrentMember();
     // GET highestRandIndexParticipatedIn
     const { data: memberStats } = useMemberStats();
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -428,11 +428,9 @@ export const useOngoingElectionData = (): UseQueryResult<
         loggedInMember?.account,
         memberStats?.ranks?.length
     );
-    console.info("useOED().membersInOngoingRound:", membersInOngoingRound);
     let { data: votingMemberData } = useMemberDataFromVoteData(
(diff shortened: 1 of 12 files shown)


brandonfancher:
Okay. I tested multi-round scenario extensively. Everything is displaying accurately! 🎉 

I will do a brief code review tomorrow morning. After that, we can :shipit: !

The remaining issues, which I can tackle in a separate PR, since loading/error states have been a focus of mine lately:
* "You aren't involved" flashes on reload even in R1
* We lose our loading state for a short period of time when loading
* Election header is incorrect with a hard-coded time
* Past rounds/state don't appear until a refetch is triggered by a window refocus or reload

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -337,48 +343,30 @@ export const getOngoingElectionData = async (
     currentElection?: CurrentElection,
     myDelegation: EdenMember[] = []
 ) => {
```
brandonfancher, line 345: `memberStats` arg here is unused

packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -50,16 +53,11 @@ export const OngoingElection = ({ election }: { election: any }) => {
...
     }
-    // console.info("currentElection:", currentElection);
-
-    const showNoConsensusInPreviousRoundMessage =
-        (myDelegationSoFar?.length || 0) < election.round;
 
     // TODO: Model all this data to be self-consistent and to abstract the frontend from the complexities of the backend logic
     // start with logged-in user
     // check that they registered for the election (and cast at least one vote?)
     // see if they're still participating
```
brandonfancher, line 60: Anything to amend or remove here? Prob not a bad idea to review comments related to this body of work throughout.

packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -104,25 +102,25 @@ export const OngoingElection = ({ election }: { election: any }) => {
...
+        <Container className="flex items-center space-x-2 pr-8 py-8">
             <BsInfoCircle
                 size={22}
                 className="ml-px text-gray-400 place-self-start mt-1"
             />
             <div className="flex-1">
                 <Text size="sm">
-                    You did not advance a delegate to participate in further
-                    rounds. Please{" "}
+                    You aren't involved in further rounds. Please{" "}
```
brandonfancher, line 123: I'd recommend "aren't" -> "are not" and, in the last sentence, "they are" -> "they will be"? Heck, we could even hide the last sentence when we're in `ElectionStatus.Final`.

packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -32,11 +34,12 @@ export const OngoingElection = ({ election }: { election: any }) => {
         isLoading: isLoadingMemberStats,
```
brandonfancher, line 34: `const { data: loggedInUser } = useCurrentMember();` can go away a few lines above this.

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
```diff
@@ -200,14 +203,7 @@ export const OngoingRoundSegment = ({
...
 
-    // TODO: this needs more attention once we have sufficient test data to test it
-    if (!participants?.length) {
-        return (
-            <div>
-                You nor a delegate of yours is participating in this round.
-            </div>
-        );
-    }
+    if (!ongoingElectionData?.isMemberStillParticipating) return null;
```
brandonfancher, line 206: This is the only place we're using `ongoingElectionData` in this component (we will use it a lot more in this component soon, but that's for later!) And we're doing >200 lines worth of queries and calculations above it only to return `null` depending on the answer to a question, which we already know. 

I recommend removing (for now) `ongoingElectionData` from the props of this component and doing this check instead inside the `CurrentRound` component that appears on line `167` of `ongoing-election.tsx`. We can short circuit a lot of this that way.

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
```diff
@@ -247,7 +243,9 @@ export const OngoingRoundSegment = ({
                         : "This round is finalizing. Please submit any outstanding votes now. You will be able to come back later to upload election videos if your video isn't ready yet."}
                 </Text>
             </Container>
-            {isVotingOpen ? (
+            {voteData?.length === 0 ? (
+                <div>you didn't participate in this round</div>
```
brandonfancher, line 247: In what event would this display? If it should _never_ display to the user, then we should consider dropping it. If it can display in some cases, let's capitalize it and wrap it properly in a `<Container />` so it presents acceptably.

brandonfancher (changes requested):
A few low-hanging, preliminary comments in case you want to process those. I'll do final review in the morn.

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
```diff
@@ -156,19 +157,21 @@ export const OngoingRoundSegment = ({
...
-    const userVoterStats = voteData.find(
+    const userVoterStats = voteData!.find(
         (vs) => vs.member === loggedInMember?.account
     );
 
-    const userVotingFor = members.find(
-        (m) => m.account === userVoterStats?.candidate
-    );
+    const userVotingFor =
+        voteData &&
```
brandonfancher, line 173: We already check for the presence of `voteData` above...I think we can drop line `173`.

packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -6,13 +6,15 @@ import {
...
 import { Button, Container, Heading, Loader, Link, Text } from "_app/ui";
 import { ErrorLoadingElection } from "elections";
-import { ActiveStateConfigType, ElectionStatus } from "elections/interfaces";
+import {
+    ActiveStateConfigType,
+    Election,
+    ElectionStatus,
+} from "elections/interfaces";
 
 import * as Ongoing from "./ongoing-election-components";
```
brandonfancher, line 20: This first TODO below can go away now.

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -413,9 +414,9 @@ export const useEncryptedData = (scope: EncryptionScope, id: string) =>
         ...queryEncryptedData(scope, id),
         enabled: Boolean(id),
     });
-export const useOngoingElectionData = (
-    queryOptions: any = {}
-): UseQueryResult<Election | undefined> => {
+export const useOngoingElectionData = (): UseQueryResult<
```
brandonfancher, line 417: Add a space between these two functions.

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -40,9 +40,11 @@ import {
     ActiveStateConfigType,
     CurrentElection,
     CurrentElection_activeState,
+    Election,
+    ElectionStatus,
```
brandonfancher, line 44: unused

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -72,11 +79,17 @@ export const getMyDelegation = async (
             ? memberStats?.ranks.length - 1
```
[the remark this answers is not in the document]
brandonfancher: We can drop the second `?` on line 79 too. We know we have `memberStats`.

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -81,12 +82,11 @@ export const getMemberGroupFromIndex = (
 export const getMemberGroupParticipants = async (
     memberAccount?: string,
     roundIndex?: number,
-    config?: ActiveStateConfigType
+    config: ActiveStateConfigType = {
+        num_participants: 13,
+        num_groups: 1,
+    }
```
brandonfancher, line 88: Should we turn this default into a named constant? It might bring clarification as to why _this_ default.

packages/webapp/src/elections/api/eden-contract.ts
```diff
@@ -293,40 +288,51 @@ const ELECTION_DEFAULTS: Election = {
...
+                    const memberData = await queryClient.fetchQuery(
+                        queryKey,
+                        queryFn
+                    );
+                    return memberData;
+                }
+            );
+            const participantsMemberData = await Promise.all(
+                pParticipantsMemberData!
+            );
```
brandonfancher, line 319: We have a mechanism for querying _all_ of this `memberData` in one network call (instead of kicking off several network calls as you are here.) Modifying this to use that mechanism will save on bandwidth and RPC calls. Take a look at how we leverage `queryMembers` within `useMemberDataFromEdenMembers` in `queries.ts` to see how to do that. I highly recommend making this change.

brandonfancher (changes requested):
Mostly minor stuff with one exception.

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
cool-ant: There's no getting around doing those 200 lines of code. They'll be gotten by 5 other components. And it's all cached. So extremely small cost, *and* given another of your comments above, I now use 2 fields from the structure, so I'd greatly prefer to leave in the nicer, more readable/maintainable structure

packages/webapp/src/elections/components/ongoing-election.tsx
```diff
@@ -104,25 +102,25 @@ export const OngoingElection = ({ election }: { election: any }) => {
...
                 className="ml-px text-gray-400 place-self-start mt-1"
             />
             <div className="flex-1">
                 <Text size="sm">
-                    You did not advance a delegate to participate in further
-                    rounds. Please{" "}
+                    You aren't involved in further rounds. Please{" "}
                     <Link href={""}>join the Community Room</Link> &amp; Support
                     for news and updates of the ongoing election. The results
                     will be displayed in the My Delegation area after the
```
brandonfancher, line 126: Let's make "My Delegation" semibold (suggestion from TH).

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round-segment.tsx
brandonfancher: I suspect we're misunderstanding one another's comments. My point is more: Why invoke this component at all if `!isMemberStillParticipating`?

brandonfancher (approved):

PR #424, opened 2021-08-20
Ensure completed rounds appear after new rounds complete

brandonfancher:
I'll squash this before I merge.

This basically ensures that all internal dependency queries inside of the `useOngoingElectionData()` query hook refetch when the current election object updates.

There are a few other mostly-related, minor tweaks herein as well.

### DIFF
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -9,6 +9,7 @@ import {
     getNewMembers,
     getMembersStats,
     MemberData,
+    MemberStats,
 } from "members";
 import { getCommunityGlobals } from "_app/api";
 
@@ -54,9 +55,12 @@ export const queryChiefDelegates = {
     queryFn: getChiefDelegates,
 };
 
-export const queryMyDelegation = (memberAccount?: string) => ({
-    queryKey: ["query_my_delegation", memberAccount],
-    queryFn: () => getMyDelegation(memberAccount),
+export const queryMyDelegation = (
+    memberStats?: MemberStats,
+    memberAccount?: string
+) => ({
+    queryKey: ["query_my_delegation", memberAccount, memberStats],
+    queryFn: () => getMyDelegation(memberAccount, memberStats),
 });
 
 export const queryMemberGroupParticipants = (
@@ -144,6 +148,7 @@ export const queryOngoingElectionData = (
     votingMemberData?: MemberData[],
     currentElection?: CurrentElection,
     myDelegation?: EdenMember[],
+    memberStats?: MemberStats,
     queryOptions: any = {}
 ) => ({
     queryKey: [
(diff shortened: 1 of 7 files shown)


packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -267,14 +269,32 @@ export const useIsCommunityActive = () => {
...
+    const { data: cachedMemberStats } = useMemberStats({
+        enabled: !memberStats,
+    });
 
-    return useQuery({
-        ...queryMyDelegation(member?.account),
-        enabled: enabled && Boolean(member?.account),
+    const stats = memberStats ?? cachedMemberStats;
+    const enabled = Boolean(member?.account && stats);
+    const highestCompletedRoundIndex = stats && stats.ranks.length - 1;
```
cool-ant, line 292: op to simplify:
const highestCompletedRoundIndex = stats?.ranks.length - 1

packages/webapp/src/_app/hooks/queries.ts
```diff
@@ -305,34 +325,30 @@ export const useCurrentElection = (queryOptions: any = {}) =>
...
 
+    let enabled = Boolean(memberAccount) && roundIndex !== undefined;
+
     return useQuery<VoteData[], Error>({
-        ...queryMemberGroupParticipants(
-            memberAccount,
-            roundIndex,
-            currentActiveElection?.config
-        ),
+        ...queryMemberGroupParticipants(memberAccount, roundIndex, config),
```
cool-ant, line 349: config still ends up being undefined if we're not in the Active state, no? gotta be a more elegant way to do this. Don't know what it is, but there's gotta be. :) no need for now, but worth noodling on. I *think* why we're doing this here is that you can say currentElection?.config due to TS. Otherwise, `currentElection?.config` would be equivalent to what's above.

cool-ant (approved):

packages/webapp/src/_app/hooks/queries.ts
brandonfancher: Yes, still `undefined` if we're in the `Active` state. But I used a type guard instead of a cast, since what comes in isn't actually always a `CurrentElection_active`.

I modified the wording the JSDoc here because, technically, the `queryFn` still returns critical data when we pass `undefined` in as `config`.

packages/webapp/src/_app/hooks/queries.ts
brandonfancher: Ah, right! The reason we can't do this here is because we follow it with an arithmetic operation. We cannot subtract `1` from `undefined`. So in this case, it _must_ be `stats && stats.ranks.length - 1`.

brandonfancher:
This is complete. I'm testing multi-round election scenario now.

My local branch has a fix to `main` that I may push here and merge if it doesn't get addressed separately before then.

I'm leaving my test election frozen in time on `test2.edev` for @cool-ant to look at to see the too-many-people-listed-in-completed-rounds issue.

brandonfancher:
My multi-round election test on `test2.edev` was a success. The problem addressed by this PR is confirmed as resolved. I'm also pushing a patch to `getTableRawRows()` to fix `main` pending a more comprehensive look at that by those who have been working on refactors to those functions.

PR #425, opened 2021-08-22
Fixing problem of members in completed rounds who aren't participating in current election, but...

cool-ant:
The enum I'm using here is obvoiusly named opposite what we'd expect, but it's used a bunch of other places. Do we know about this value changing or why this seems inverted?

### DIFF
--- a/packages/webapp/src/elections/api/eden-contract.ts
+++ b/packages/webapp/src/elections/api/eden-contract.ts
@@ -9,6 +9,7 @@ import {
     CONTRACT_MEMBER_TABLE,
     CONTRACT_VOTE_TABLE,
     didReachConsensus,
+    ElectionParticipationStatus,
     getRow,
     getTableRawRows,
     getTableRows,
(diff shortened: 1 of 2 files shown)


brandonfancher:
I think the enum is correct, not flipped. If you’re logged in as edenmember12, you’ll notice that all the extras are still participating in round 2. So your code is simply filtering those out and that’s why you get the right list. 

(IIRC, participation status in the members table is updated for a user when their role in the current election ends.)

But, theres still a problem: That won't filter out people in the member table who participated in a prior election and voted for edenmember12 but are not participating the _current_ election.

Clearing the representative values from the members table for ALL members (participating or not) at the beginning of a new election would fix this. (There may be other, better approaches too,)

brandonfancher:
> But, theres still a problem: That won't filter out people in the member table who participated in a prior election and voted for edenmember12 but are not participating the current election.

Ah, I see! (I was confusing presence-in-vote-table with participation status). Yes, this explicitly _does_ filter those people out, because they're not participating in the current election.

Great!

Still don't understand what you mean about the inverted enum value.

brandonfancher (approved):

PR #426, opened 2021-08-23
Add Delegate Funds Available & Withdraw Component

sparkplug0025:
Add a delegate funds banner on top of the profile page.

image

- Withdraw button is displayed only if the logged in user = the profile account
- The withdraw action will sum all the disbursements, transfer the funds to the eden member account balance and then withdraw it to the eos account.

### DIFF
--- a/packages/webapp/src/delegates/api/eden-contract.ts
+++ b/packages/webapp/src/delegates/api/eden-contract.ts
@@ -3,12 +3,14 @@ import { EdenMember } from "members";
 import { queryClient } from "pages/_app";
 import {
     CONTRACT_DISTRIBUTION_ACCOUNTS_TABLE,
-    getTableIndexRows,
+    getTableRows,
     i128BoundsForAccount,
+    INDEX_BY_OWNER,
     isValidDelegate,
     queryElectionState,
     queryMemberByAccountName,
     queryVoteDataRow,
+    TABLE_INDEXES,
 } from "_app";
 
 import { DistributionAccount } from "../interfaces";
@@ -77,10 +79,6 @@ export const getMyDelegation = async (
     return myDelegates;
 };
 
-// eosio secondary indexes for distaccount table defined at:
-// /contracts/eden/include/distributions.hpp
-const INDEX_BY_OWNER = 2;
-
 export const getDistributionsForAccount = async (
     account: string
 ): Promise<DistributionAccount[]> => {
@@ -90,13 +88,16 @@ export const getDistributionsForAccount = async (
         return DISTRIBUTION_ACCOUNTS;
     }
 
-    const distributionRows = await getTableIndexRows(
+    const distributionRows = await getTableRows(
         CONTRACT_DISTRIBUTION_ACCOUNTS_TABLE,
-        INDEX_BY_OWNER,
-        "i128",
-        lower,
-        upper,
-        9999
+        {
+            ...TABLE_INDEXES[CONTRACT_DISTRIBUTION_ACCOUNTS_TABLE][
+                INDEX_BY_OWNER
+            ],
+            lowerBound: lower,
+            upperBound: upper,
+            limit: 9999,
+        }
     );
 
     return distributionRows as DistributionAccount[];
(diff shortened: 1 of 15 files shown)


packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -69,3 +76,28 @@ export const getMyDelegation = async (
 
     return myDelegates;
 };
+
+// eosio secondary indexes for distaccount table defined at:
+// /contracts/eden/include/distributions.hpp
+const INDEX_BY_OWNER = 2;
```
cool-ant, line 82: This should be in the TABLE_INDEXES definition in _app/eos/api.ts

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -69,3 +76,28 @@ export const getMyDelegation = async (
...
+    const { lower, upper } = i128BoundsForAccount(account);
+
+    if (devUseFixtureData) {
+        return DISTRIBUTION_ACCOUNTS;
+    }
+
+    const distributionRows = await getTableIndexRows(
+        CONTRACT_DISTRIBUTION_ACCOUNTS_TABLE,
+        INDEX_BY_OWNER,
+        "i128",
```
cool-ant, line 96: Yeah, and this can now use TABLE_INDEXES as used in elections/api/eden-contracts.ts L120

packages/webapp/src/delegates/components/delegate-funds-available.tsx
```diff
@@ -0,0 +1,97 @@
...
+              distributions.map((distribution) => distribution.balance)
+          )
+        : undefined;
+
+    // omit component if it's not a current delegate and available funds are empty
+    if ((!member || !member.election_rank) && !availableFunds) {
+        return null;
+    }
+
+    const nextDisbursementDate = "October 15, 2021"; // TODO: how to calculate this? create fn.
```
cool-ant, line 37: Let's add a TODO in the code at minimum just to track this as an open item

packages/webapp/src/delegates/components/delegate-funds-available.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const availableFunds = distributions
+        ? sumAssetStrings(
+              distributions.map((distribution) => distribution.balance)
+          )
+        : undefined;
+
+    // omit component if it's not a current delegate and available funds are empty
+    if ((!member || !member.election_rank) && !availableFunds) {
+        return null;
+    }
```
brandonfancher, line 35: A member could be allocated funds shortly before a new election and not withdraw them in time. We need to allow any active member (even if they're no longer a delegate) to withdraw when they have funds available to them in one or more distribution accounts.

packages/webapp/src/delegates/components/delegate-funds-available.tsx
brandonfancher: I'd prefer that we either figure out how to calculate this _in this PR_ OR omit the "Check back on x..." sentence entirely and only expose it once it's accurate. I'm getting wary of merging hard-coded (and inaccurate) placeholder information this close to launching our next milestone.

packages/webapp/src/delegates/components/delegate-funds-available.tsx
```diff
@@ -0,0 +1,97 @@
...
+                        >
+                            Withdraw
+                        </Button>
+                    )}
+                </div>
+            </div>
+            {ualAccount?.accountName === member?.account && (
+                <Text>
+                    Delegate funds are disbursed monthly. Check back on{" "}
+                    {nextDisbursementDate} for your next disbursement.
```
brandonfancher, line 92: See comment above.

brandonfancher (changes requested):

packages/webapp/src/delegates/components/delegate-funds-available.tsx
sparkplug0025: Thats exactly why I check available funds as well... if the member has not an election rank we will show the widget because it has available funds. Does it make sense?

brandonfancher: Awesome! I missed that. Thanks!

brandonfancher (approved):

PR #427, opened 2021-08-23
Fix: getting my delegation can result in infinite loop

brandonfancher:
I just ran an election that uncovered this bug. Here's the setting:

- Election in Round 2
- `edenmember12` is delegate participating in Round 2
- `edenmember12` was represented previously by `edenmember11` and so `edenmember11` is the value for `representative` in the `member` table entry for `edenmember12`
- `edenmember11` participated in the current election's Round 1 with `edenmember12`.
- `getMyDelegation` looks for `edenmember12`'s R1 participants, finds `edenmember11`. Then looks at `edenmember11`'s delegate, finds `edenmember12`...and so on.

This fix ensures that the loop will stop after completing for all the rounds.

### DIFF
--- a/packages/webapp/src/delegates/api/eden-contract.ts
+++ b/packages/webapp/src/delegates/api/eden-contract.ts
@@ -43,6 +43,7 @@ export const getMyDelegation = async (
 
     let nextMemberAccount = loggedInMemberAccount;
     let isHeadChief: Boolean;
+    let currRound = highestCompletedRoundIndex + 1;
     do {
         let member = await getMemberWrapper(nextMemberAccount);
         if (!member)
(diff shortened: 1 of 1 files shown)


cool-ant (dismissed):
Yes, good call. My original My Delegation code had something just like this in it before I simplified My Delegation to *not* calculate a Delegation until the election was over. Now that we're back to having it calculate a Delegation *during* an election, this fits the bill.

packages/webapp/src/delegates/api/eden-contract.ts
```diff
@@ -65,7 +66,8 @@ export const getMyDelegation = async (
         }
         isHeadChief = member.account === member.representative;
         nextMemberAccount = member.representative;
-    } while (isValidDelegate(nextMemberAccount) && !isHeadChief);
+        currRound -= 1;
```
cool-ant, line 69: You need to decrement this in the for loop as well to ensure you run out of rounds appropriately. That for loop will add members to the Delegation, which means it should reduce the currRound as well.

brandonfancher: "as well" or "instead of?"

brandonfancher: Got it. Thanks!

cool-ant (approved):

PR #429, opened 2021-08-23
Election loading and error state improvements

brandonfancher:
- Shores back up loading and error state handling in the UI after the addition of the `useOngoingElectionData()` hook.
- Handles previously-unhandled seeding and post-round election statuses.
- Removes inaccurate election end time estimates (for now).
- Makes vote button loading text more accurate and (related) differentiates the checkbox next to a participant depending on whether you've actually voted for them (grey) or have only selected them to cast a vote for them (blue).

### DIFF
--- a/packages/webapp/src/elections/components/ongoing-election-components/voting-round-participants.tsx
+++ b/packages/webapp/src/elections/components/ongoing-election-components/voting-round-participants.tsx
@@ -57,7 +57,7 @@ const VotingRoundParticipants = ({
                                 isSelected={
                                     selectedMember?.account === member.account
                                 }
-                                userIsVotingFor={
+                                hasCurrentMembersVote={
                                     member.account === userVotingFor
                                 }
                                 onSelect={() => selectMember(member)}
--- a/packages/webapp/src/elections/components/ongoing-election-components/voting-round-participants.tsx
+++ b/packages/webapp/src/elections/components/ongoing-election-components/voting-round-participants.tsx
@@ -9,6 +9,7 @@ interface VotingRoundParticipantsProps {
     voteData: VoteData[];
     selectedMember: MemberData | null;
     onSelectMember: (member: MemberData) => void;
+    userVotingFor?: string;
 }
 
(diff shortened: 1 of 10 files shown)


packages/webapp/src/pages/election/components.tsx
```diff
@@ -47,14 +47,17 @@ export const MembersPage = (props: Props) => {
...
                         member={member}
                         isSelected={selectedMember === member.account}
                         onSelect={() => setSelected(member.account)}
-                        votesReceived={member.account === "pip.edev" ? 5 : 0}
-                        votingFor="Philip Pip"
+                        votesReceived={
+                            member.account === "edenmember12" ? 5 : 0
+                        }
+                        votingFor="Test Member 5"
+                        userIsVotingFor={member.account === "edenmember13"}
```
cool-ant, line 60: I assume all this is hard-coded because we don't have this data for some reason?
Please at least add a TODO here so we don't lose track of it, but I'd prefer this test data came from somewhere else, like a fixture.

cool-ant (approved):
Solid cleanup.
A few minor comments; main was is the test data you have hard-coded.

packages/webapp/src/elections/components/ongoing-election-components/voting-round-participants.tsx
```diff
@@ -55,6 +57,9 @@ const VotingRoundParticipants = ({
                                 isSelected={
                                     selectedMember?.account === member.account
                                 }
+                                userIsVotingFor={
+                                    member.account === userVotingFor
+                                }
```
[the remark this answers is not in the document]
brandonfancher: yeah, I had a hard time naming that. Open to suggestions. It's answering the question: is the logged in member voting for this participant in the participants list?

packages/webapp/src/pages/election/components.tsx
brandonfancher: Yeah, this is all just a dummy components display that's probably already out of date in other ways. Good call, though; I'll add a TODO on that.

PR #433, opened 2021-08-24
Add next disbursement info

sparkplug0025:
- Add a query for the distribution state
- Adds an informational text about the next distribution or the actual distribution state underneath the withdraw banner

### DIFF
--- a/packages/webapp/src/delegates/components/delegate-funds-available.tsx
+++ b/packages/webapp/src/delegates/components/delegate-funds-available.tsx
@@ -1,4 +1,6 @@
+import dayjs from "dayjs";
 import { useState } from "react";
+
 import {
     assetToString,
     Button,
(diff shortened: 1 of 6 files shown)


packages/webapp/src/delegates/components/delegate-funds-available.tsx
```diff
@@ -84,6 +91,42 @@ export const DelegateFundsAvailable = ({ account }: Props) => {
...
+                    ).format("LLL")}{" "}
+                    for your next disbursement.
+                </Text>
+            );
+        case "election_distribution":
+            return (
+                <Text>
+                    An election is running and a disbursement will be made as
+                    soon as the election ends.
+                </Text>
```
brandonfancher, line 120: I recommend we tweak the language here to reflect that they only get this next disbursement if they come out of the ongoing election as a delegate:

> An election is currently underway. Disbursements to newly-elected delegates will be processed as soon as the election ends.

brandonfancher (approved):
This looks great! I'll make the tweak I suggested to the text, commit, and merge.

PR #434, opened 2021-08-24
Participation and related password detection/update tweaks

brandonfancher:


packages/webapp/src/elections/components/registration-election-components/participation-card.tsx
```diff
@@ -315,13 +319,18 @@ const ConfirmParticipationModal = ({ isOpen, close }: ModalProps) => {
                     onSubmit={submitParticipationConfirmation}
                     isLoading={isLoading}
                     onCancel={close}
+                    deadline={deadline}
                 />
             )}
+            {step === ParticipationStep.ConfirmParticipationSuccess && (
+                <ConfirmParticipationStepSuccess close={close} />
+            )}
```
sparkplug0025, line 327: ❤️

packages/webapp/src/encryption/components/reenter-password-form.tsx
```diff
@@ -59,13 +64,23 @@ export const ReenterPasswordForm = ({
...
                 </Form.LabeledSet>
                 <div className="flex space-x-3">
                     <Button type="neutral" onClick={onCancel}>
                         Cancel
                     </Button>
-                    <Button isSubmit disabled={!fields.password || isLoading}>
+                    <Button
+                        isSubmit
+                        disabled={!fields.password || isLoading}
+                        isLoading={isLoading}
```
sparkplug0025, line 82: **future-enhancement (not blocking for this pr):** I feel like the isLoading flag should automatically disable the button in the button component. I can't think of a case where a process is loading but we want the button to be able to be pressed again... I think that would remove some redundancy in our code like this:
`<Button isSubmit disabled={!fields.password} isLoading={isLoading} />`

brandonfancher: Yeah, that's crossed my mind too. It makes sense.

sparkplug0025 (approved):
Looks great! I noticed that in the forgot password messages you reworded to not ask to copy/paste/store in a safe place. Was it intended?

brandonfancher:
> Looks great! I noticed that in the forgot password messages you reworded to not ask to copy/paste/store in a safe place. Was it intended?

Thanks! And no, good catch. I'll make sure that wording is included before I merge. 🙏

PR #437, opened 2021-08-25
Add token balance and history button in profile

sparkplug0025:
Adds a token balance component with a history link in the member profile page.
image


packages/webapp/src/_app/api/eosio.token-contract.ts
```diff
@@ -0,0 +1,19 @@
...
+export const getTokenBalanceForAccount = async (
+    account: string
+): Promise<Asset> => {
+    const data = await eosJsonRpc.get_currency_balance(
+        "eosio.token",
+        account,
+        minimumDonationAmount.symbol
+    );
+
+    if (data && data.length) {
```
cool-ant, line 14: data?.length

cool-ant (approved):

PR #438, opened 2021-08-25
Add to Calendar button

sparkplug0025:
image


thomas-hallgren:
@sparkplug0025 Here is the UX for that link: https://xd.adobe.com/view/06e9c2f7-3fb6-4274-95ce-4ae4c8d8fff6-53ba/
Screen Shot 2021-08-25 at 12 45 28 PM

PR #441, opened 2021-08-26
Election round video upload

cool-ant:
This PR still has
- No buttons in Election screen that launch this page in a new window

To Do:
- Test in election run-through
- Test refetching of ongoingElectionData to update how many round show (as rounds complete and start)
- No reason to show ongoing round until voting is possible (avoid the "starts in..." period)


sparkplug0025:
closing in favor of #508

PR #447, opened 2021-08-29
New layout: sidebar on desktop, tab bar on mobile

brandonfancher:
## Tasks
- [x] Tablet/Desktop: Sidebar navigation
- [x] Mobile: Header and bottom tab nav
- [x] Reflow Community page for new nav
- [x] Reflow Home page for new nav
- [x] Reflow Member detail page
- [x] Reflow Membership page
- [x] Reflow Induction detail page
- [x] Profile icon image/fallback icon top left on mobile, otherwise in sidebar
- [x] Sign out button
- [x] Reflow Zoom Oauth page
- [x] Add footer back in at bottom of pages...potentially just with links
- [ ] Improve home page layout

## Future PRs / Ideas
I have to make UI adjustments to these pages anyway, so I'll knock these out when I do that.
- Reflow Delegation page
- Reflow Election page (during ongoing election)
- Make `<EncryptionPasswordAlert />` sticky to top
- On wider screens, footer links in side bar?
- Make sure we're not wasting valuable left-hand real estate at various sub-full-desktop breakpoints
- Make sidebar scroll when not enough vertical space


packages/webapp/src/_app/ui/mobile-nav.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    className="flex flex-col justify-center items-center space-y-1"
+                >
+                    {children}
+                </a>
+            </Link>
+        </div>
+    );
+};
+
+const AccountMenu = () => {
```
sparkplug0025, line 114: suggestion: maybe we want to extract this component due to the signout and signin functionalities (to avoid having bugs where signing in / signing out works in mobile but not in desktop - or vice-versa). So we could reuse it here and in the header-nav. `<AccountMenu type="mobile">` --> then inside you have the same HoC wrapping two components: `MobileAccountMenu` and `DesktopAccountMenu`.

packages/webapp/src/_app/ui/side-nav.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <HeaderLogo />
+            <HeaderItems menuItems={MENU_ITEMS} />
+            <div className="-mr-2.5 xl:mr-0">
+                <NavProfile location="side-nav" />
+            </div>
+        </nav>
+    </header>
+);
+
+const HeaderLogo = () => (
```
sparkplug0025, line 25: suggestion (future prs, perhaps?): I noticed that we use the Logo in various ways... Maybe we should centralize all of them in a Logo file and reuse from there. (I imagine probably getting some DRYness in the process)

packages/webapp/src/_app/ui/table.tsx
```diff
@@ -34,13 +34,13 @@ export const Table = ({
     rowClassName = "",
     tableHeader,
 }: TableProps) => {
-    const tableClass = `xs:m-0 sm:-mx-4 md:m-0 bg-white border-t border-b md:border border-gray-200 md:rounded md:shadow-sm text-gray-700 ${className}`;
+    const tableClass = `bg-white border-t border-b border-gray-200 text-gray-700 ${className}`;
     return (
-        <div className="mb-8">
+        <div className="">
```
sparkplug0025, line 39: chore: remove classname

packages/webapp/src/_app/config.tsx
```diff
@@ -0,0 +1,68 @@
...
+        href: "/election",
+        label: "Election",
+        NavIcon: () => (
+            <FiCheckSquare
+                className="text-3xl xl:text-2xl -ml-px -mt-px -mb-px"
+                style={{ fontSize: 28 }}
+            />
+        ),
+    },
+};
```
sparkplug0025, line 68: thought: I don't like the idea of mixing react components with plain configs... When I imagine configs I think about declaring stuff... so the old router could be extended as:
```ts
{ 
  INDUCTION: {
    href: "/induction",
    label: "Membership",
    icon: AiOutlineIdCard,
    size: "regular",
  }
}
```

I see that we have some where they have images, so the interface could offer icon or image as optional... And yeah, there are some variations  of styles and padding, but I think we could wrap them in parameters too.

I think we are totally fine for now but we should probably revisit at some point.

packages/webapp/src/pages/index.tsx
```diff
@@ -155,58 +128,106 @@ const ElectionCard = () => {
...
+                    >
+                        {memberStats.pending_members} pending invitation
+                        {memberStats.pending_members !== 1 && "s"}
+                    </Link>
+                </>
+            )}
         </Card>
-    ) : null;
+    );
+};
```
sparkplug0025, line 231: thought: I just realized how big this page is. I think we should keep following the pattern of having nextjs `pages` lean, only responsible for routing and prefetching and have the components in their corresponding domain package. (as we are doing with elections, inductions, treasury etc.)

sparkplug0025 (commented):
Super duper awesome job!  👏  👏  👏  👏 
I love this revamp. And I love even more the way that while you worked on this you actually refactored and simplified a lot of components!

Aside from the comments, I realized that the Zoom Oauth page was not included in the Sidebar layout. Did you forget or was it in purpose?

packages/webapp/src/_app/layouts/side-nav-layout.tsx
```diff
@@ -0,0 +1,32 @@
...
+            <title>{title && `${title} | `} Eden</title>
+        </Head>
+        <SideNav />
+        <MobileNav />
+        <div className="flex-1 flex flex-col w-full mt-14 xs:mt-0 pb-16 xs:pb-0">
+            {banner}
+            <main className={`min-h-screen border-r ${className}`}>
+                {children}
+            </main>
+            {/* <Footer /> */}
```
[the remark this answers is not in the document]
brandonfancher: Planning to tackle footer re-introduction today. And as for the Zoom Oauth page, thanks for catching that. I forgot about it!

packages/webapp/src/_app/config.tsx
brandonfancher: Yeah, agreed. I felt some dissonance here too. I'll lodge an issue for us to revisit this.

packages/webapp/src/pages/index.tsx
brandonfancher: Yes, it is too big. I'll break it apart as part of my Home page work I'm finishing today. And thank you for the review!!!

packages/webapp/src/_app/config.tsx
brandonfancher: Fixed. The problem I was working around was obviated by some new SVGs Thomas just sent me. 🤘

packages/webapp/src/_app/ui/mobile-nav.tsx
brandonfancher: I'll turn this into an issue to address post-v0.2 launch.

packages/webapp/src/_app/ui/side-nav.tsx
brandonfancher: I'll turn this into an issue to address post-v0.2 launch.

packages/webapp/src/pages/index.tsx
brandonfancher: Will tackle shortly.

packages/webapp/src/_app/config.ts
```diff
@@ -1,12 +1,52 @@
...
+    },
+    TREASURY: {
+        href: "/treasury",
+        label: "Treasury",
+        Icon: TreasuryNav,
+    },
+    ELECTION: {
+        href: "/election",
+        label: "Election",
+        Icon: ElectionNav,
```
cool-ant, line 50: nice

packages/webapp/src/_app/utils/navigation.ts
```diff
@@ -0,0 +1,5 @@
+import { ROUTES } from "_app/config";
+
+export const NAV_MENU_ITEMS = Object.keys(ROUTES)
+    .map((k) => ROUTES[k])
+    .filter((k) => !k.hideNav);
```
cool-ant, line 5: nice

cool-ant (approved):

PR #455, opened 2021-09-01
Password Improvements

sparkplug0025:
- derives encryption keys from 12 characters passwords
- enforce password store management handling


brandonfancher (approved):
Tested and LGTM!

PR #465, opened 2021-09-03
Add Image UI element with Fallback property

sparkplug0025:


brandonfancher (approved):
looks great!

brandonfancher:
I added 8 fallback avatar images (SVGs) that Thomas put together for us. I'm using only one for now--arguably the most generic. In a future PR, maybe we can randomize among the 8.

PR #470, opened 2021-09-07
Add election participants counter

sparkplug0025:
image

### DIFF
--- a/contracts/eden/src/eden-micro-chain.cpp
+++ b/contracts/eden/src/eden-micro-chain.cpp
@@ -884,9 +884,7 @@ void handle_event(const eden::election_event_schedule& event)
 
 void handle_event(const eden::election_event_begin& event)
 {
-   db.elections.emplace([&](auto& election) { 
-      election.time = event.election_time; 
-   });
+   db.elections.emplace([&](auto& election) { election.time = event.election_time; });
 }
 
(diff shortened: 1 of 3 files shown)


libraries/debug_eos_vm/include/debug_eos_vm/dwarf.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <eosio/stream.hpp>
 #include <memory>
+#include <optional>
```
sparkplug0025, line 5: @tbfleming without this line clang was failing. is it fine to add this change?

tbfleming: yes

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -157,6 +157,7 @@ struct status
    std::string memo;
    eosio::block_timestamp nextElection;
    uint16_t electionThreshold = 0;
+   uint16_t electionParticipants = 0;
```
tbfleming, line 160: Please use `numElectionParticipants`. This allows us to add `electionParticipants` in the future if needed to list the participants.

sparkplug0025: done.

tbfleming (approved):

PR #473, opened 2021-09-08
Add election participation count query

sparkplug0025:
- add the first box subchain query! 🎉
- shows the total number of participants on the current election
- if the query fails we simply don't show the component (it means that we don't depend on the box -- I think we should adopt this pattern whenever possible, so other eden communities are not forced to spin up a box with history... I also understand that this might be a deal breaker for some functionalities that depends on nice massaged data from history... hence why "whenever possible" -- thoughts?)
- it standardizes the `useQuery` from `ReactSubchain` with the same interface as the `react-query`, then we have a nice object to control when it's loading, has an error or has the query data result in fact.

### DIFF
--- a/packages/common/src/subchain/ReactSubchain.tsx
+++ b/packages/common/src/subchain/ReactSubchain.tsx
@@ -45,7 +45,14 @@ export function useCreateEdenChain(
 
 export const EdenChainContext = createContext<SubchainClient | null>(null);
 
-export function useQuery<T = any>(query: string): T {
+export interface Query<T> {
+    isLoading: boolean; // flags if the query is loading
+    data?: T; // has the query data when it's loaded or empty if not found
+    error?: any; // flags an error in case it fails to load the query
+}
+
+export function useQuery<T = any>(query: string): Query<T> {
+    console.info("using query...");
     const client = useContext(EdenChainContext);
     const [cachedQuery, setCachedQuery] = useState<string | null>();
     // non-signalling state
@@ -53,7 +60,7 @@ export function useQuery<T = any>(query: string): T {
         mounted: true,
         cachedClient: null as SubchainClient | null,
         subscribed: null as SubchainClient | null,
-        cachedQueryResult: null as any,
+        cachedQueryResult: { isLoading: false } as Query<T>,
     });
     useEffect(() => {
         return () => {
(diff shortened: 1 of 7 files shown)


packages/common/src/subchain/ReactSubchain.tsx
```diff
@@ -75,9 +81,15 @@ export function useQuery<T = any>(query: string): T {
...
         if (client?.subchain) {
-            state.cachedQueryResult = client.subchain.query(query);
+            state.cachedQueryResult = {
+                ...client.subchain.query(query),
+                isLoading: false,
+            };
         } else {
-            state.cachedQueryResult = null;
+            state.cachedQueryResult = {
+                isLoading: false,
```
tbfleming, line 90: true; `client?.subchain` switches to non-null after wasm and state are loaded

tbfleming:
Approach looks good

brandonfancher:
> if the query fails we simply don't show the component (it means that we don't depend on the box -- I think we should adopt this pattern whenever possible, so other eden communities are not forced to spin up a box with history... I also understand that this might be a deal breaker for some functionalities that depends on nice massaged data from history... hence why "whenever possible" -- thoughts?)

I'm pretty sure we're going all-in on box. At least, the way I've been envisioning it, all RPC queries will be replaced with references to browser-side derived state in the not-too-distant future. 

Other Eden communities will have a few options:
* spin up their own box (totally do-able)
* use a business-in-a-box provider
* rely on our box until BIAB providers are available

In this case, I think it's totally fine that we simply show nothing on error. But we'll want fallback error UI in the future as we lean more and more on box.

packages/common/src/subchain/ReactSubchain.tsx
```diff
@@ -45,15 +45,21 @@ export function useCreateEdenChain(
 
 export const EdenChainContext = createContext<SubchainClient | null>(null);
 
-export function useQuery<T = any>(query: string): T {
+export interface Query<T> {
+    isLoading: boolean; // flags if the query is loading
+    data?: T; // has the query data when it's loaded or empty if not found
+    errors?: any; // flags an error in case it fails to load the query
+}
```
brandonfancher, line 52: ❤️

brandonfancher (commented):
This looks really good. 🎉 

I'd like to test it tomorrow morning with you. If/when you're available, hit me up on Telegram. I having trouble successfully testing this. I got a DFUSE API key, updated my wasm file, removed the old json file and spun it up. And I got it to display the total _for a moment_ but then it errored out. Haven't had a chance to investigate this evening. If you can help me spin it up and test tomorrow morn, I'll approve so we can merge.

packages/common/src/subchain/ReactSubchain.tsx
sparkplug0025: ah okay, so I will remove the error here

brandonfancher (approved):
Love it!

PR #482, opened 2021-09-08
Testnet-only UAL softkey authenticator

brandonfancher:
## Tasks
- [x] Scaffold and get it prompting for login, account name, etc.
- [x] Implement login/logout with localstorage.
- [x] Key validation against public keys on account...wrong password message to user if not correct.
- [x] Implement signing via eosjs. 

Resource: https://github.com/EOSIO/ual-eosio-reference-authenticator

## Goals
- GOOD: User in dev can choose Softkey UAL Authenticator. After they enter their account name, UAL calls up (probably via our context reducer and the `useUALSoftkeyLogin` hook) one of our modals asking them to enter their "password". The user enters a private key, it's validated, added to localstorage, and they're in. Signing requests pull it from localstorage at sign time to sign with it.
- BETTER: The password field also accepts an election-password-style password (same algorithm elections use) and attempts to get the private key out of it to sign with. IF it's an election-password-style password, it also assumes that that is the user's election password and adds it to that entry in localstorage as well.

Why?

Because I can then give the Zoom testers 1 password and they won't be faced with having to create a second password for elections and have to remember which to use in which context.


packages/webapp/.env
```diff
@@ -22,7 +22,7 @@ NEXT_PUBLIC_SUBCHAIN_WS_URL = "ws://localhost:3032/v1/subchain/eden-microchain"
 NEXT_PUBLIC_SUBCHAIN_SLOW_MO = "false"
 NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT = "10.00000000 WAX"
 # NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT = "10.0000 EOS"
-NEXT_PUBLIC_ENABLED_WALLETS = "ANCHOR,LEDGER"
+NEXT_PUBLIC_ENABLED_WALLETS = "ANCHOR,LEDGER,SOFTKEY"
```
sparkplug0025, line 25: Should we remove Softkey here or leave it as is to help dev and always remember to override in prod?

packages/webapp/src/_app/eos/ual/softkey/softkey-user.ts
```diff
@@ -0,0 +1,191 @@
...
+        privateKey: string
+    ): Promise<void> {
+        try {
+            const edenMember = await getEdenMember(this.accountName);
+            const publicKey = PrivateKey.fromString(privateKey)
+                .getPublicKey()
+                .toLegacyString();
+            if (publicKey === edenMember?.encryption_key) {
+                // key already set on account; just set as encryption password in browser
+                this.loginHook.updateEncryptionPassword(publicKey, privateKey);
```
sparkplug0025, line 86: Good catch!! 👍

packages/webapp/src/_app/eos/ual/softkey/ual-softkey-login-modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    const doSubmit = (e: React.FormEvent) => {
+        e.preventDefault();
+
+        let password: string;
+        let incorrectPassword =
+            "PVT_K1_4nHj22pv211y1CqgaAJLmkPxS688Ha19kPzT2bxLNXMcN7aCb";
+        if (fields.password.length < 12) {
+            password = incorrectPassword; // trigger incorrect password message
```
sparkplug0025, line 35: I'm not following this. Why not simply throw "password must be at least 12 chars"? What does having this default PK here means?

sparkplug0025 (approved):
just a question before merging

packages/webapp/src/_app/eos/ual/softkey/ual-softkey-login-modal.tsx
brandonfancher: There's nothing to catch a throw here for this submit handler. So we get an unhandled runtime error. Feeding it a password we know to be incorrect triggers the regular incorrect-password error handling so I figured that was good enough.

packages/webapp/.env
brandonfancher: We shouldn't have to override in prod, as there will be a different chainId and so `SOFTKEY` should come back as not available.

sparkplug0025: oh, great!

packages/webapp/src/_app/eos/ual/softkey/ual-softkey-login-modal.tsx
brandonfancher: `onError` FTW. Thanks!

PR #492, opened 2021-09-15
Init Induction E2E Tests

sparkplug0025:
Starting to have super cool tests that make sense :-)

### DIFF
--- a/packages/webapp/cypress/support/commands.ts
+++ b/packages/webapp/cypress/support/commands.ts
@@ -24,18 +24,35 @@
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 
+/**
+ * Logins with UAL using Eden SoftKey mode
+ */
 Cypress.Commands.add("login", (account) => {
     cy.get(`[data-testid="signin-nav-buttonsm"]`).first().click();
-    cy.wait(500);
     cy.get("#ual-box div").contains("Password").click();
-    cy.wait(500);
     cy.get('#ual-box input[type="text"]').type(account);
-    cy.wait(500);
     cy.get("#ual-box div").contains("Continue").click();
-    cy.wait(500);
     cy.get('input[type="password"]').type(Cypress.env("test_users_pk"));
     cy.get('button[type="submit"]').click();
-    cy.wait(500);
+});
+
+/**
+ * Intercept EOS RPC API Calls
+ */
+Cypress.Commands.add("interceptEosApis", () => {
+    cy.intercept("**/chain/get_table_rows").as("eosGetTableRows");
+    cy.intercept("**/chain/push_transaction").as("eosPushTransaction");
+});
+
+/**
+ * This is a wrapper around the `cy.wait()` command for now; since adding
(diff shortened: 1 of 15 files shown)


cool-ant (approved):

packages/webapp/cypress/support/commands.ts
```diff
@@ -24,4 +24,18 @@
...
+    cy.wait(500);
+    cy.get("#ual-box div").contains("Password").click();
+    cy.wait(500);
+    cy.get('#ual-box input[type="text"]').type(account);
+    cy.wait(500);
+    cy.get("#ual-box div").contains("Continue").click();
+    cy.wait(500);
+    cy.get('input[type="password"]').type(Cypress.env("test_users_pk"));
+    cy.get('button[type="submit"]').click();
+    cy.wait(500);
```
brandonfancher, line 38: I remember seeing documentation about `wait` being an antipattern sometimes in the Cypress documentation. Just found it: https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

They also talk specifically in that section about alternatives to `cy.wait(arbitrary_time_period)` when used with `cy.get(...)`. 

I don't know whether the way you're doing it here is warranted or is the antipattern. But thought I'd mention it just in case.

packages/webapp/cypress/support/index.ts
```diff
@@ -14,7 +14,14 @@
...
 
-// Alternatively you can use CommonJS syntax:
-// require('./commands')
+declare namespace Cypress {
+    interface Chainable {
+        /**
+         * Custom command to login with a specified account
+         * @example cy.login('alice.edev')
+         */
+        login(account: string): Chainable;
```
brandonfancher, line 25: nice!

packages/webapp/cypress/support/commands.ts
sparkplug0025: you are absolutely right when the waiting is unnecessary... I consciously added in two situations:
1. handle ual inputs, because it seemed like chained `cy.get()` didn't work...
2. wait for blocks to propagate after creating the induction, and this one was needed only for our CI pipeline because the headless browser runs super fast... 

That being said, let me go through this cypress reference to see if I have better ways of handling it!

sparkplug0025: Adjusted! Can you please review it again?

packages/webapp/cypress/support/commands.ts
```diff
@@ -24,18 +24,35 @@
...
+
+/**
+ * This is a wrapper around the `cy.wait()` command for now; since adding
+ * `cy.wait()` should almost never be used, we actually have a genuine case
+ * for that since we need to wait for blocks to read some data from the chain.
+ * Also it makes our tests free of undesirable waits, if we have a new legit
+ * case for `cy.wait()` we can just add a new command here.
+ */
+Cypress.Commands.add("waitForBlocksPropagation", (blocks = 3) => {
+    cy.wait(blocks * 500);
```
brandonfancher, line 55: Nice!

brandonfancher (approved):
🎉

PR #507, opened 2021-09-18
retry dfuse after receiving 'complete'

tbfleming:
The dfuse client normally retries automatically when the connection goes down. Unfortunately, it sometimes doesn't retry when a `complete` message arrives.


packages/box/src/dfuse-receiver.ts
```diff
@@ -83,6 +83,7 @@ async function webSocketFactory(
 
 export default class DfuseReceiver {
     storage: Storage;
+    stream: Stream | null = null;
```
sparkplug0025, line 86: question: why not `undefined`? it plays better with typescript

sparkplug0025 (approved):
lgtm, codewise. Aside from the suggestion on the undefined, I'd also suggest to add a PR summary to help us understanding what is going on. I'm assuming it's a Dfuse connection retry when the `complete` message is received, but I'm not sure why... a little bit of context would be super nice!

packages/box/src/dfuse-receiver.ts
tbfleming: I got in the habit of using null since json and graphql don't have undefined

PR #509, opened 2021-09-20
Add induction inviter/witness names to induction detail pages

brandonfancher:
Listing witnesses on induction/invite tables kicked off way too many RPC calls, causing RPC endpoints to block requesting IPs for 10 minutes at a time (so many requests are perceived as a potential DDoS attack). In previous PRs, we removed witness lists from these tables altogether. But inviters and witnesses do depend on this information to organize their groups, and it quickly becomes confusing for them when they have too many open inductions to keeps straight.

This PR adds the inviter/witness information back, but to the induction DETAIL pages. When the information is fetched for one induction at a time, the number of requests is smaller (vs. fetching this info x 10 for a list of inductions.) Furthermore, we're already getting some of the information required to display this data on the induction detail pages, further reducing additional requests required.

image

image

image

### DIFF
--- a/packages/webapp/src/inductions/hooks.ts
+++ b/packages/webapp/src/inductions/hooks.ts
@@ -1,9 +1,15 @@
 import { useMemo } from "react";
 import { useQuery } from "react-query";
 
-import { queryInductionWithEndorsements, useUALAccount } from "_app";
+import {
+    queryInductionWithEndorsements,
+    useMemberByAccountName,
+    useMemberListByAccountNames,
+    useUALAccount,
+} from "_app";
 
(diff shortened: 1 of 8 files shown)


packages/webapp/src/inductions/hooks.ts
```diff
@@ -24,3 +30,28 @@ export const useGetInductionWithEndorsements = (inductionId: string) =>
         // gets rid of bugs cause by refetching right after anchor/scatter signing
         refetchOnWindowFocus: false,
     });
+
+export const useInductionParticipants = (inductionId: string) => {
```
sparkplug0025, line 34: praise: nice hook!

PR #518, opened 2021-09-23
Use election voting data query for the Election Upload Service

sparkplug0025:
Uses Eden box solution to actually query the past votes and uploaded videos.

### DIFF
--- a/packages/webapp/src/_app/hooks/box-queries.ts
+++ b/packages/webapp/src/_app/hooks/box-queries.ts
@@ -1,4 +1,6 @@
-import { useQuery } from "@edenos/common/dist/subchain";
+import { Query, useQuery } from "@edenos/common/dist/subchain";
+import dayjs from "dayjs";
+import { MemberAccountData } from "members";
 
 export interface ElectionStatusQuery {
     status: {
@@ -18,3 +20,111 @@ export const useElectionStatus = () =>
     }
 }
 `);
+
+export interface CurrentMemberElectionVotingDataQuery {
+    electionTime: dayjs.Dayjs;
+    votes: [
+        {
+            roundIndex: number;
+            votingBegin: dayjs.Dayjs;
+            votingEnd: dayjs.Dayjs;
+            votingStarted: boolean;
+            votingFinished: boolean;
+            resultsAvailable: boolean;
+            candidate?: MemberAccountData;
+            winner?: MemberAccountData;
+            video: string;
+        }
+    ];
+}
+
+export const useCurrentMemberElectionVotingData = (
+    account?: string
+): Query<CurrentMemberElectionVotingDataQuery> => {
+    const query = useQuery<any>(
+        account
+            ? `
+    {
+      members(le: "${account}", ge: "${account}") {
+        edges {
+          node {
+            account
+            elections(last: 1) {
+              edges {
+                node {
+                  time
+                  votes {
+                    edges {
+                      node {
+                        group {
+                          round {
+                            round
+                            votingBegin
+                            votingEnd
+                            votingStarted
+                            votingFinished
+                            resultsAvailable
+                          }
+                          winner {
+                            account
+                            profile {
+                              name
+                            }
+                          }
+                        }
+                        candidate {
+                          account
+                          profile {
+                            name
+                          }
+                        }
+                        video
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    `
+            : ""
+    );
+
+    let data = query.data;
+    if (data) {
+        const lastElection =
+            data.members.edges?.[0]?.node.elections?.edges?.[0].node;
+        console.info(lastElection);
+        if (lastElection) {
+            data.electionTime = dayjs(lastElection.time);
+            data.votes =
+                lastElection.votes?.edges?.map((voteEdge: any) => ({
+                    roundIndex: voteEdge.node.group.round.round,
+                    votingBegin: dayjs(voteEdge.node.group.round.votingBegin),
+                    votingEnd: dayjs(voteEdge.node.group.round.votingEnd),
+                    votingStarted: voteEdge.node.group.round.votingStarted,
+                    votingFinished: voteEdge.node.group.round.votingFinished,
+                    resultsAvailable:
+                        voteEdge.node.group.round.resultsAvailable,
+                    candidate: voteEdge.node.candidate
+                        ? {
+                              account: voteEdge.node.candidate.account,
+                              name: voteEdge.node.candidate.profile.name,
+                          }
+                        : undefined,
+                    winner: voteEdge.node.group.winner
+                        ? {
+                              account: voteEdge.node.group.winner.account,
+                              name: voteEdge.node.group.winner.profile.name,
+                          }
+                        : undefined,
+                    video: voteEdge.node.video,
+                })) || [];
+        }
+    }
+
+    return { ...query, data };
+};
--- a/packages/webapp/src/_app/hooks/box-queries.ts
+++ b/packages/webapp/src/_app/hooks/box-queries.ts
@@ -93,14 +93,12 @@ export const useCurrentMemberElectionVotingData = (
             : ""
     );
 
-    let data = query.data;
-    if (data) {
+    if (query.data) {
         const lastElection =
-            data.members.edges?.[0]?.node.elections?.edges?.[0].node;
-        console.info(lastElection);
+            query.data.members.edges?.[0]?.node.elections?.edges?.[0].node;
         if (lastElection) {
-            data.electionTime = dayjs(lastElection.time);
-            data.votes =
+            query.data.electionTime = dayjs(lastElection.time);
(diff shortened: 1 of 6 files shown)


packages/webapp/src/_app/hooks/box-queries.ts
```diff
@@ -18,3 +20,109 @@ export const useElectionStatus = () =>
...
+                    winner: voteEdge.node.group.winner
+                        ? {
+                              account: voteEdge.node.group.winner.account,
+                              name: voteEdge.node.group.winner.profile.name,
+                          }
+                        : undefined,
+                    video: voteEdge.node.video,
+                })) || [];
+        }
+    }
```
sparkplug0025, line 125: question: any suggestions on how to make this function more elegant? I see us repeating the pattern of massaging/formatting retrieved box queries... Question is, how to set an elegant pattern so we don't grow this kind of code?

packages/webapp/src/pages/election/round-video-upload.tsx
```diff
@@ -24,19 +22,21 @@ import {
...
 
 export const RoundVideoUploadPage = () => {
     const [isElectionComplete, setIsElectionComplete] = useState(false);
-    const [uploadLimitTime, setUploadLimitTime] = useState(dayjs());
+    const [uploadLimitTime, setUploadLimitTime] = useState(
+        dayjs().add(1, "day") // set default limit in the future while election data is being loaded
+    );
 
     const {
         data: currentElection,
```
brandonfancher, line 42: Re: Line 36 - line 65:

suggestion: I think this is a case where we can reactively derive state instead of setting state. Could we get rid of both `useState` and `useEffect` calls here with something like this instead:

```
    const {
        data: currentElection,
        isLoading: isLoadingElection,
        isError: isErrorElection,
    } = useCurrentElection();
    const { data: electionState } = useElectionState();
    
    let uploadTimeLimit = electionState
        ? dayjs(electionState.last_election_time + "Z").add(48, "hour")
        : dayjs().add(1, "day");

    const isAfterUploadLimitTime =
        dayjs().isAfter(uploadTimeLimit) &&
        currentElection?.electionState === ElectionStatus.Registration;
```

Then we in the JSX on `isAfterUploadLimitTime`.

brandonfancher: Also, should we react to `isLoading` and `isError` from `useElectionState()` like we do from `useCurrentElection()`?

packages/webapp/src/_app/hooks/box-queries.ts
brandonfancher: That's a good question. I'll give it some thought.

suggestion: this is just a thought--I'm not totally committed to it--but our queries file for react-query queries became super long because it was all in one file in `_app`. With our new box queries, should we consider putting them inside the directories their related to? So instead of this query going in `_app/hooks/box-queries.ts`, it would be in `election/hooks/box-queries.ts`? Of course, any common queries could still go in the former.

packages/webapp/src/pages/election/round-video-upload.tsx
```diff
@@ -112,47 +112,27 @@ export default RoundVideoUploadPage;
...
         return <ErrorLoadingElection />;
     }
 
     if (
-        !ongoingElectionData ||
-        (ongoingElectionData.completedRounds.length === 0 &&
-            ongoingElectionData.ongoingRound.participantsMemberData.length ===
-                0)
+        !currentMemberElectionVotingData ||
+        !currentMemberElectionVotingData.votes.length
```
brandonfancher, line 135: suggestion: `!currentMemberElectionVotingData?.votes.length`

packages/webapp/src/pages/election/round-video-upload.tsx
```diff
@@ -196,43 +169,44 @@ const RoundVideoUploadList = ({ election }: { election: CurrentElection }) => {
...
-            setRoundVideoList(roundVideoList);
-
             setVideoSubmissionPhase(undefined);
-            setUploadCompleteMessage("Election video uploaded successfully!");
+            setUploadCompleteMessage({
+                roundIndex,
+                message: "Election video uploaded successfully!",
+            });
         } catch (error) {
             onError(error, "Unable to set the election round video");
```
brandonfancher, line 178: question: my IDE has started complaining to me about these errors:

image

Any idea why? It stops complaining if I do: `onError(error as Error, "Unable to set the ele...")`

Is that happening for you too?

packages/webapp/src/pages/election/round-video-upload.tsx
```diff
@@ -326,10 +257,10 @@ const Header = ({
     roundEndTime,
 }: HeaderProps) => {
     const subText =
-        roundStartTime && roundEndTime
-            ? `${roundStartTime.format("LT")} - ${roundEndTime.format("LT z")}`
-            : winner && isValidDelegate(winner.account)
+        winner && isValidDelegate(winner.account)
             ? `Delegate elect: ${winner.name}`
```
brandonfancher, line 261: suggestion: remove the word "elect" here, since this can appear in the context of a completed election where they're officially a delegate now. ("delegate election" or "president elect" or "congressman elect" usually refer to someone who has been elected but has not yet taken office.)

brandonfancher (changes requested):
Here's my code review. On to functional review!

packages/webapp/src/pages/election/round-video-upload.tsx
```diff
@@ -112,47 +112,27 @@ export default RoundVideoUploadPage;
 const RoundVideoUploadList = ({ election }: { election: CurrentElection }) => {
```
brandonfancher, line 112: suggestion: remove the unused `election` prop

brandonfancher:
suggestion: I still think we need to handle non-logged-in user and non-member states.

This is what we get if you're not logged in:

image

And this is what we get if you're logged in as a non-Eden-member:

image

The first says there was an error when there actually isn't (it should show a sign in button) and the second doesn't handle a real error (it should show the "Join Eden" button sort of like the the participation card:

image

packages/webapp/src/pages/election/round-video-upload.tsx
```diff
@@ -161,13 +141,6 @@ const RoundVideoUploadList = ({ election }: { election: CurrentElection }) => {
             </Text>
```
brandonfancher, line 141: suggestion: 

```jsx
<Container>
    <Text>
        It appears you haven't participated in any election round.
        Nothing to upload.
    </Text>
</Container>
```

This will fix the padding issues.

brandonfancher:
suggestion: wording

Let's change this text to:

> Election video upload

>The election video upload service will open in a new browser tab. It is available during the 48 hours after the beginning of the election. Any election videos must be uploaded during this period.

> Due to the large size of election videos, we recommend uploading them from a desktop computer whenever possible.

image

packages/webapp/src/_app/hooks/box-queries.ts
cool-ant: I want to say something like tagged templates. If we could store the query in a string (get them out of this code), and delay the interpolation, that would be much cleaner.
The way them seem structured makes it doable, but with some hacking...
https://betterprogramming.pub/javascript-how-backticks-work-de269e0fb8ba
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

brandonfancher:
I just tested this with a three-level election in all sorts of ways. It's working great!

Once these suggestions are addressed, this is shippable and usable. 👍 

i love it!

packages/webapp/src/pages/election/round-video-upload.tsx
sparkplug0025: yes, done!

packages/webapp/src/pages/election/round-video-upload.tsx
sparkplug0025: Man, I started with a fresh environment today and I started to receive this!

sparkplug0025: (It was not happening before)

sparkplug0025:
> I created a new MemberGateContainer component. 

Every time we need to show sign in / join eden we can reuse it. It's tricky in the participation button because there are more logic involved there... but this will be enough here and to new features too.

packages/webapp/src/_app/hooks/box-queries.ts
sparkplug0025: I like both suggestions. Since it's about just splitting and moving around I'm going to keep like this in this PR and we can add these improvements to our backlog!

brandonfancher:
> > suggestion: I still think we need to handle non-logged-in user and non-member states.
> 
> I created a new MemberGateContainer component.
> 
> Every time we need to show sign in / join eden we can reuse it. It's tricky in the participation button because there are more logic involved there... but this will be enough here and to new features too.

praise: great addition!

brandonfancher (approved):
Awesome!

PR #520, opened 2021-09-24
My delegates: styles and states

brandonfancher:
- [ ] Shore up styles
- [ ] Handle loading and error states
- [ ] Refactor / break out components
- [ ] Test with multi-level election


brandonfancher:
I’ll look into the type error tomorrow morning.

packages/webapp/src/delegates/components/my-delegation.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+export default MyDelegation;
+
+const LowerDelegates = ({ members, myDelegation }: Props) => {
+    const { data: membersStats, isLoading, isError } = useMemberStats();
+
+    if (isLoading) return <LoadingContainer />;
+    if (isError || !membersStats) return <ErrorLoadingDelegation />;
+
+    const heightOfDelegationWithoutChiefs = membersStats.ranks.length - 2;
```
cool-ant, line 43: This assumes a post-election state. Either that or it won't consider the top 2 levels prior to end of election, no?

packages/webapp/src/delegates/components/my-delegation.tsx
```diff
@@ -0,0 +1,174 @@
...
+        return <ErrorLoadingDelegation />;
+    }
+
+    return (
+        <>
+            <div>
+                <LevelHeading className="border-b">
+                    Chief Delegates
+                </LevelHeading>
+                <MembersGrid members={memberData}>
```
cool-ant, line 150: Ok just to clarify: Chief Delegates will show all delegates including the Head Chief. and
Will the logged-in user's rep have a delegate badge in the UI so the user can see who their delegate is?

packages/webapp/src/delegates/components/my-delegation.tsx
brandonfancher: Right. Results on the "My Delegation" page do not show up during an election.

packages/webapp/src/pages/delegates/index.tsx
```diff
@@ -1,220 +1,100 @@
...
+                {isLoading ? (
+                    <LoadingContainer />
+                ) : isError ? (
+                    <ErrorLoadingDelegation />
+                ) : isElectionInProgress ? (
+                    <ElectionInProgress />
+                ) : !myDelegation || !myDelegationMemberData ? (
+                    <NoDelegationToDisplay />
+                ) : (
+                    <MemberGateContainer>
```
cool-ant, line 88: Member + Delegate ==> MemberGate?

packages/webapp/src/delegates/components/my-delegation.tsx
brandonfancher: For now, all Chief Delegates have a delegate badge. So there's room for improvement here. There shouldn't be any changes to logic on this page in this PR, though. I'm aiming to maintain the status quo.

cool-ant: It looks like this is a pretty strict refactor / adding of very limited items like Loading and Error states, so... I'll just consider my comments on this comments-to-self. :)

cool-ant (approved):

packages/webapp/src/pages/delegates/index.tsx
brandonfancher: I'm not following.

Issue #510, opened 2021-09-20
eden: [election] global election view driven by Todd's blocks-to-browser state engine


brandonfancher:
This is a read-only, publicly-available view into an ongoing election. The UI can be found here: https://xd.adobe.com/view/ef42e383-6bc6-414e-9290-7d6dd6e27313-ba11/screen/5ceccade-0cec-48b5-b76a-38abaa13f9aa/

We need to:

1. Hook up to the blocks-to-browser state engine to make all the data required for this view available to the page/component, and
1. Take the existing components and use them in this view.

brandonfancher:
@sparkplug0025 I sent you some feedback via Telegram. FYI.

PR #523, opened 2021-09-27
Add election stats query

sparkplug0025:
Creates the Election Stats page as per #510 

image

Also tested with mocked videos:
image

### DIFF
--- a/packages/webapp/src/_app/hooks/box-queries.ts
+++ b/packages/webapp/src/_app/hooks/box-queries.ts
@@ -69,6 +69,7 @@ export const useCurrentMemberElectionVotingData = (
                             account
                             profile {
                               name
+                              img
                             }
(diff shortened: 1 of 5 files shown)


cool-ant (approved):
I'd love to see the consensometer in the <GroupHeader />, but I know that's bonus at this point...

sparkplug0025:
> I'd love to see the consensometer in the , but I know that's bonus at this point...

I'm not adding the consesometer since there's no ui space for that, we should double check with Thomas to see if there's a place to fit this somewhere in the current screen... Nonetheless, that was a good suggestion because I inspected the consensometer code and I actually realized that I forgot to tally the votes and show the lead candidates. So now I'm reusing the `tallyVotesFromVoteData` to show the leads!

Issue #243, opened 2021-05-25
eden: rpc endpoint load balancing and failover


brandonfancher:
When the primary RPC node isn't responding, fallback to another. We should have multiple RPC endpoints in our env vars and config. And we should also consider randomizing the order in which they're used.

React Query allows you to tap into their retry logic--globally and on a query-by-query basis: https://react-query.tanstack.com/guides/query-retries

That might be a good place to start.

Requirements:

- When you first visit the page, the RPC provider you're using should be randomly selected. (load balancing)
- If an RPC call fails for one reason or another (including you kicked off too many requests and got put in purgatory for 10 minutes), the next RPC endpoint should be promoted and used (until there's a problem with that one or everything is re-randomized when you reload the web app)

sparkplug0025:
I think we need more thought on this... The fallback for the queries will work fine when we add custom retries on React-Query but since we are not using mutations for broadcasting the transactions we will not have the same behavior when signing transactions... 

I think we could do that by wrapping our transactions broadcasting after signing it with a react-query mutation.

cool-ant:
@sparkplug0025 here are the production endpoints I've now vetted. Please use these
eos.dfuse.eosnation.io
eos.greymass.com
api.eossweden.org
mainnet.eosamsterdam.net

PR #527, opened 2021-09-29
Implement the EOS JSON RPC with Balancer

sparkplug0025:
Implements requirements from: #243 

- Explicitly creates an Env Var for all the Reads RPCs: `NEXT_PUBLIC_EOS_READ_RPC_URLS`. That way we don't need to worry about UAL not being able to dynamically swap RPCs and the user will probably not get locked out from the writing RPC just by submitting transactions as he/she uses Eden as we expect for normal usage.
- Meat is on [class EosJsonRpcWithBalancer](https://github.com/eoscommunity/Eden/compare/sparkplug0025/243-rpc-fallback?expand=1#diff-8581022c8cb8f11e5552bc1a152692d62cf3dd85409e414a724cfe5c6faa3ec5):
  - We initialize the balancer selecting a random RPC
  - When fetching data from the RPC we parse the error messages
  - If it's a locking error message we rotate to the next provided RPC

**DANGER! CAUTION! WARNING! >>> Set the env config var `NEXT_PUBLIC_EOS_READ_RPC_URLS` when deploying to production!!!**


brandonfancher:
This is super valuable and definitely merge-worthy as it stands. I do have a couple observations though:

1. One of the risks we'd like to mitigate here is that an RPC endpoint simply goes down. Maybe we start getting 500s or any other failure code. It doesn't look like--in the code or in my testing--that we cycle if, say, we start getting a 500 or 404, etc. Now, a user can still refresh and re-roll the dice. And that's certainly better than nothing! But it'd be nice if we cycled endpoints for these other failure scenarios too.
2. Our write RPC host will likely be included as one of the EOS_READ_RPC_URLS too. If that one fails and the user cycles over to another endpoint, reads will work, but they won't be able to push any transactions. Any way we can easily make the write/transaction-push scenarios simply use the current endpoint from the `EosJsonRpcWithBalancer` instance? (Even if a failure to push doesn't cause it to cycle over yet?)

I agree with doing this outside of/apart from React Query's retry mechanism for now. 👍

brandonfancher (approved):
This is super valuable and definitely merge-worthy as it stands. I do have a couple observations though:

1. One of the risks we'd like to mitigate here is that an RPC endpoint simply goes down. Maybe we start getting 500s or any other failure code. It doesn't look like--in the code or in my testing--that we cycle if, say, we start getting a 500 or 404, etc. Now, a user can still refresh and re-roll the dice. And that's certainly better than nothing! But it'd be nice if we cycled endpoints for these other failure scenarios too.
2. Our write RPC host will likely be included as one of the EOS_READ_RPC_URLS too. If that one fails and the user cycles over to another endpoint, reads will work, but they won't be able to push any transactions. Any way we can easily make the write/transaction-push scenarios simply use the current endpoint from the `EosJsonRpcWithBalancer` instance? (Even if a failure to push doesn't cause it to cycle over yet?)

I agree with doing this outside of/apart from React Query's retry mechanism for now. 👍

tbfleming (approved):

brandonfancher:
@sparkplug0025 Given the comments, I'll leave it up to you whether you want to go ahead and merge and make suggested changes in a follow-up PR, or append them to this PR.

We deployed this branch today for our dev test and we didn't have any trouble at all with network requests / fetches.

PR #529, opened 2021-09-30
Zoom only eat cookies

sparkplug0025:
🍪

This PR addresses the Zoom Security Review concerns.

For that we moved our Zoom Account JWT to a cookie as opposed of using localstorage -- the rationale is that JS code cannot have access to Zoom JWT tokens; hence, a cookie with `httpOnly` is being used.

### DIFF
--- a/packages/webapp/package.json
+++ b/packages/webapp/package.json
@@ -24,6 +24,7 @@
         "ipfs-http-client": "^50.0.0",
         "ipfs-only-hash": "^4.0.0",
         "next": "~10",
+        "nookies": "^2.5.2",
         "nprogress": "^0.2.0",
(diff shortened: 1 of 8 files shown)


brandonfancher (approved):
The code looks good. I cannot test this locally; we'll merge into `main` and I will test against the deployed dev environment.

packages/webapp/package.json
```diff
@@ -24,6 +24,7 @@
         "ipfs-http-client": "^50.0.0",
         "ipfs-only-hash": "^4.0.0",
         "next": "~10",
+        "nookies": "^2.5.2",
```
tbfleming, line 27: hahahaha

PR #538, opened 2021-10-01
See app as another user

tbfleming:


packages/webapp/src/pages/as/[id].tsx
```diff
@@ -0,0 +1,80 @@
...
+    const ualContext = useContext<any>(UALContext);
+    if (ualContext) {
+        ualContext.activeUser = new UnloggedinUser(account);
+        useRouter().push(ROUTES.ELECTION.href);
+    }
+
+    return (
+        <SingleColLayout title="Loading...">
+            <LoadingCard />
+        </SingleColLayout>
```
Comment (brandonfancher) [inline packages/webapp/src/pages/as/[id].tsx:76]: nit: change to:

```
<SideNavLayout title="Loading...">
    <LoadingContainer />
</SideNavLayout>
```

Otherwise, we see the old layout while loading and it's kind of jarring.

brandonfancher (approved):
This. Is. Awesome.

👏

packages/webapp/src/pages/as/[id].tsx
Comment (tbfleming) [reply packages/webapp/src/pages/as/[id].tsx]: changed

PR #554, opened 2021-10-07
Throttle outgoing webapp RPC get_table_row fetches

brandonfancher:
In my testing today in dev, I was able to create the circumstances during an election that challenged (and sometimes crossed) the limits acceptable to our RPC providers. This wasn't a problem during polling; it was really only so on refresh / page load / or when going from a normal round to a chief delegates round with 11 members. I could get it to hit as high as ~32 fetches in a second.

Wow, it's going to be nice when we refactor this on top of the box state derivation engine / wasm!

Anyway, I got one RPC endpoint to block me. And that's in dev. And that's not good.

IMHO, this represents a real threat to the stability of our real election on Saturday. And it's worth fixing in this late hour.

We've built RPC endpoint failover, and that's great. But it's not bulletproof by any means, and a user trying to load a page with lots of delegates on it could get blocked by all of them. No good.

Then it dawned on me that we could throttle fetches (the offenders are all `get_table_rows`) on our end. So I found this gem: https://github.com/vercel/async-sema.

With a few lines of code, we can feed it a per second rate limit and it will do the rest.

So far, my tests are very good. Everything is stable and fetch throttling is happening as it should.


cool-ant (approved):
solid

brandonfancher:
> Another solution to this issue is to balance queries across a few different endpoints. We balance queries across over a dozen endpoints for eospowerup.io and you can see the reliability metrics here: https://stats.boid.com/d/X2YaUPRnk/eospowerup?orgId=1&from=now-7d&to=now&kiosk=tv&fullscreen&panelId=26

That's a really cool interface! Thank you for sharing your approach to addressing this problem. Do you have a link to the code that does the load balancing?

We implemented something similar in #527 in that if we encounter a fetch error with one endpoint, we roll over to another RPC endpoint. That has provided _some_ resiliency, but that doesn't really solve this specific problem. Many endpoints limit or block you (for 10 min) if you send too many requests in a second. And that's what we're running into. And it wouldn't take much to burn _all_ the RPC bridges.

Since the election is so close, this is a simple stopgap that should give us the reliability we're looking for. (I'm definitely willing to sacrifice a little speed for reliability.) If we were going to continue relying so heavily on RPC `get_table_rows` for our data, I'd say we should consider implementing your solution. But I think we're going to refactor much of this to depend on @tbfleming's blocks-to-browser client-side state derivation solution after this first election. In fact, we're already using it this election for our global election results/stats page, video upload page, and more. So this will be a good first test for it.

Thanks again for chiming in, @jdheeter. This is what the EOS community is all about! ❤️

PR #569, opened 2021-10-11
Completes induction with file uploads

sparkplug0025:
- Adds a complete induction test using our ephemeral chain.
- Removes the unnecessary Box E2E test since we now have a full environment on our ephemeral chain:
  - testing a real chain with contracts
  - testing a real box
  - testing a real webapp
  - hence, renamed from WebApp E2E Tests to simply E2E Tests


brandonfancher (approved):
This looks great to me!

PR #572, opened 2021-10-12
Community members from box queries

brandonfancher:


packages/webapp/cypress/support/commands.ts
```diff
@@ -29,7 +29,9 @@
  */
 Cypress.Commands.add("login", (account) => {
     cy.get(`[data-testid="signin-nav-buttonsm"]`).first().click();
+    cy.wait(500);
     cy.get("#ual-box div").contains("Password").click();
+    cy.wait(500);
```
brandonfancher, line 34: For some reason, I've had quite a few test failures on the induction tests reporting that `#ual-box input[type="text"]` can't be found. These waits seem to have alleviated that, but I'm still not quite happy with this.

brandonfancher:
Actually, if we set `NODE_ENV=test` on our CICD build, it will read from the `.env.test` file. (See [NextJS | Test Environment Variables](https://nextjs.org/docs/basic-features/environment-variables#test-environment-variables.))

In that way, we can keep all this cicd contract/box stuff in `.env.test` and reserve `.env` for something else. But should that "something else" be `localhost` + `test.edev`? That's probably closest to the status quo and will require us to override less in our `.env.local` file.

brandonfancher:
> Actually, if we set `NODE_ENV=test` on our CICD build, it will read from the `.env.test` file. (See [NextJS | Test Environment Variables](https://nextjs.org/docs/basic-features/environment-variables#test-environment-variables.))

I just demonstrated this in #578. (I'm having a hard time getting tests to pass in that PR, though...due to the same issues I seem to have smoothed out in this PR!)

@sparkplug0025 If you like the approach in #578, I'll either merge that and rebase this one, or bring the changes from #578 into this PR (if I can't get tests to pass in that PR).

packages/webapp/cypress/support/commands.ts
sparkplug0025: yeah I had these waits in the past exactly because of that... I removed trusting the cy.get would wait and detect it as they should... but still... flakey...

packages/webapp/src/_app/hooks/box-queries.ts
```diff
@@ -1,7 +1,57 @@
...
+                formatQueriedMemberAccountData(member.node)
+            );
+        }
+    }
+
+    return {
+        ...pagedResult,
+        result: { ...pagedResult.result, data: formattedMembers },
+    };
+};
```
sparkplug0025, line 54: suggestion (non-blocking/next-prs): I think since we are starting to add more queries we should think about that idea of splitting the queries into their own domain folder. same for the regular queries.ts

packages/webapp/src/pages/members/index.tsx
```diff
@@ -9,24 +9,22 @@ import {
...
 
 export const getServerSideProps: GetServerSideProps = async ({ query }) => {
     const queryClient = new QueryClient();
 
-    const membersPage = parseInt((query.membersPage as string) || "1");
     const newMembersPage = parseInt((query.newMembersPage as string) || "1");
 
     await Promise.all([
-        queryClient.prefetchQuery(queryMembersStats),
-        queryClient.prefetchQuery(queryMembers(membersPage, MEMBERS_PAGE_SIZE)),
```
sparkplug0025, line 29: question: we actually never leveraged our server side rendering mechanism to generate seo tags, page titles, nice google pages previews etc. should we refactor all of our pages to not rely on server side rendering anymore? do we want the community to be discoverable?

sparkplug0025 (approved):
LGTM! 

> I just demonstrated this in #578. (I'm having a hard time getting tests to pass in that PR, though...due to the same issues I seem to have smoothed out in this PR!)
>
> @sparkplug0025 If you like the approach in #578, I'll either merge that and rebase this one, or bring the changes from #578 into this PR (if I can't get tests to pass in that PR).

Yep, that will be way better!

packages/webapp/src/_app/hooks/box-queries.ts
brandonfancher: 💯 I have that as a future action item in my description, actually.

packages/webapp/src/pages/members/index.tsx
brandonfancher: I think we _do_ want the community to be discoverable, but I think that we will de facto remove the SSR stuff as we move to box subchain data and away from RPC calls. We can still leverage it in the future for SEO though. (as we discussed)

packages/webapp/cypress/support/commands.ts
brandonfancher: Looks like I need to keep these around

PR #591, opened 2021-10-27
Init SHiP receiver

sparkplug0025:
Integrates the blocks from the SHIP plugin and add it to the microchain!  

To enable SHIP in a local chain (as opposed of using DFuse), set the following env vars:
```
SUBCHAIN_RECEIVER = "SHIP"
SHIP_ADDRESS = "127.0.0.1"
SHIP_PORT = "8080"
```

You must have a nodeos running with the eden contract deployed under account `eden.gm`. Instructions can be found here: https://github.com/eoscommunity/Eden/pull/587


packages/box/src/handlers/subchain.ts
```diff
@@ -152,8 +151,22 @@ export async function startSubchain() {
...
+
+const setupReceiver = async () => {
+    switch (subchainConfig.receiver) {
+        case SubchainReceivers.DFUSE: {
+            const dfuseReceiver = new DfuseReceiver(storage);
+            return dfuseReceiver.start();
+        }
+        case SubchainReceivers.SHIP: {
+            const shipReceiver = new ShipReceiver(storage);
+            return shipReceiver.start();
```
tbfleming, line 169: await

contracts/eden/tests/include/nodeos-runner.hpp
```diff
@@ -25,10 +25,15 @@ struct nodeos_runner
       eosio::execute("nodeos -d " + runner_name +
                      " "                                       //
                      "--config-dir config "                    //
+                     "--plugin eosio::chain_plugin "           //
```
tbfleming, line 28: chain_plugin is a default plugin

contracts/eden/tests/include/nodeos-runner.hpp
```diff
@@ -25,10 +25,15 @@ struct nodeos_runner
...
                      "--plugin eosio::chain_api_plugin "       //
                      "--plugin eosio::producer_api_plugin "    //
                      "--plugin eosio::state_history_plugin "   //
+                     "--plugin eosio::http_plugin "            //
```
tbfleming, line 32: http_plugin is automatically pulled in by *_api_plugin

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1909,6 +1982,53 @@ bool add_block(subchain::block&& eden_block, uint32_t eosio_irreversible)
    return add_block(std::move(block), eosio_irreversible);
 }
 
+[[clang::export_name("getShipBlocksRequest")]] bool getShipBlocksRequest()
+{
+   auto head = block_log.head();
+   auto block_num = head ? head->num : 1;
+
+   eosio::ship_protocol::request request = eosio::ship_protocol::get_blocks_request_v0{
+       .start_block_num = block_num,
```
tbfleming, line 1991: ship needs the eosio block number

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1869,6 +1923,37 @@ bool add_block(subchain::block&& eden_block, uint32_t eosio_irreversible)
...
+               uint32_t eosio_irreversible,
+               eosio::block_timestamp timestamp,
+               std::vector<eosio::ship_protocol::transaction_trace> traces)
+{
+   subchain::eosio_block eosio_block;
+   eosio_block.num = block.block_num;
+   eosio_block.id = block.block_id;
+   eosio_block.previous = prev.block_id;
+   eosio_block.timestamp = timestamp.to_time_point();
+   eosio_block.transactions = std::move(ship_to_eden_transactions(traces));
```
tbfleming, line 1953: std::move not needed on returned values

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1823,6 +1824,59 @@ void filter_block(const subchain::eosio_block& block)
...
+                   .name = act_trace->act.name,
+                   .creatorAction = creatorAction,
+                   .hexData = std::move(hexData),
+               };
+
+               transaction.actions.push_back(action);
+            }
+         }
+
+         transactions.push_back(transaction);
```
tbfleming, line 1873: `std::move(transaction)`

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1823,6 +1824,59 @@ void filter_block(const subchain::eosio_block& block)
...
+                      .seq = act_trace->creator_action_ordinal,
+                      .receiver = creator_action_trace.receiver,
+                  };
+               }
+
+               std::vector<char> data(act_trace->act.data.pos, act_trace->act.data.end);
+               eosio::bytes hexData{data};
+
+               subchain::action action{
+                   .seq = act_trace->action_ordinal,
```
tbfleming, line 1861: `act_trace->receipt.global_sequence`

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1823,6 +1824,59 @@ void filter_block(const subchain::eosio_block& block)
...
+            if (auto* act_trace = std::get_if<eosio::ship_protocol::action_trace_v0>(&action_trace))
+            {
+               std::optional<subchain::creator_action> creatorAction;
+               if (act_trace->creator_action_ordinal.value > 0)
+               {
+                  const auto& creator_action_trace =
+                      std::get<eosio::ship_protocol::action_trace_v0>(
+                          trx_trace->action_traces[act_trace->creator_action_ordinal.value - 1]);
+                  creatorAction = subchain::creator_action{
+                      .seq = act_trace->creator_action_ordinal,
```
tbfleming, line 1852: Use `creator_action_ordinal` to get the creator's trace, then from that, `->receipt.global_sequence`

packages/box/src/handlers/subchain.ts
```diff
@@ -152,8 +151,22 @@ export async function startSubchain() {
...
     } catch (e: any) {
         logger.error(e);
     }
 }
+
+const setupReceiver = async () => {
+    switch (subchainConfig.receiver) {
+        case SubchainReceivers.DFUSE: {
+            const dfuseReceiver = new DfuseReceiver(storage);
+            return dfuseReceiver.start();
```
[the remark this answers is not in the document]
sparkplug0025: actually we dont need await here, since we are returning a promise it's the responsibility of the caller to use `await setupReceiver()`

tbfleming:
The initial ship request still isn't correct. This use case (transient test chains) won't trip a problem case, so we can leave it as-is for now.

If we ever need to move ship support into production:
* The wasm will need to filter out unneeded blocks
* The initial block must be either:
  * One block past last irreversible (does a partial replay), or
  * Some point after that block, but with `have_positions` filled in for each eosio block after irreversible
* One of:
  * Verify we don't hit the web socket overrun case (this problem is unique to js websocket libraries), or
  * Switch from the ws library to one which implements a stream interface to avoid overrun, or
  * Change `max_messages_in_flight` to a small number and implement ack

We don't have to deal with overrun now because the transient eosio chain is small.

tbfleming (approved):

PR #595, opened 2021-11-02
Ephemeral CI

sparkplug0025:
Uses the Ephemeral chain in our CI E2E tests


.github/workflows/build.yml
```diff
@@ -356,7 +356,7 @@ jobs:
                   context: .
 
     webapp-e2e:
-        needs: build-micro-chain
+        needs: [build-cpp, build-micro-chain]
```
tbfleming, line 359: build-cpp doesn't always run when webapp-e2e needs to run. Probably need to expand what build-micro-chain builds and publish the runners from that.

.github/workflows/build.yml
```diff
@@ -392,14 +392,37 @@ jobs:
...
                   path: build
 
+            - name: Download Ephemeral Chain Runners
+              if: steps.filter.outputs.src == 'true'
+              uses: actions/download-artifact@v2
+              with:
+                  name: Ephemeral Eden Chains Runners
+                  path: build
+
+            - name: Download clsdk
```
tbfleming, line 402: This may get tricky. Optimal solution, if it's possible with GH actions:
* If build-cpp builds, then grab the clsdk artifact
* If build-cpp doesn't build, then grab the clsdk from somewhere else. The question is, where?

tbfleming (approved):

PR #597, opened 2021-11-02
Withdraw Options

brandonfancher:
## In this PR
Enables a user to specify a withdraw-to account name, amount and memo.

## Remaining
- [x] Refactor into appropriate components
- [x] Subchain refactors
- [x] Run it through the tests
- [x] Make sure new env vars get where they need to go (before merge)


packages/webapp/.env
```diff
@@ -15,6 +15,8 @@ NEXT_PUBLIC_TABLE_ROWS_MAX_FETCH_PER_SEC = "10"
 NEXT_PUBLIC_EDEN_CONTRACT_ACCOUNT = "test.edev"
 NEXT_PUBLIC_AA_FETCH_AFTER="1633883520000"
 NEXT_PUBLIC_TOKEN_CONTRACT = "eosio.token"
+NEXT_PUBLIC_TOKEN_SYMBOL = "WAX"
+NEXT_PUBLIC_TOKEN_PRECISION = "8"
```
sparkplug0025, line 19: suggestion: this is redundant because we derive the symbol and the precision from the `NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT`.

brandonfancher: I agree it's redundant. But I don't like the idea of deriving this info from `NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT` now that we're using it for more than minimum donation amount.

I propose we do it the other way around: make `NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT` simply a number (`NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT ="10"`) and derive the asset from that using the `NEXT_PUBLIC_TOKEN_SYMBOL` and `NEXT_PUBLIC_TOKEN_PRECISION`.

That seems more explicit to me now. Thoughts?

packages/webapp/src/members/components/funds-available-cta.tsx
```diff
@@ -0,0 +1,93 @@
...
+                        Eden funds available
+                    </Heading>
+                    <Heading size={4} className="block xs:hidden">
+                        Funds available
+                    </Heading>
+                    <Text>
+                        {availableFunds
+                            ? assetToLocaleString(
+                                  availableFunds,
+                                  tokenConfig.precision
```
sparkplug0025, line 61: Question: Are big numbers being displayed properly in small screens?

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
```diff
@@ -0,0 +1,150 @@
...
+                                    {availableFunds?.symbol}
+                                </p>
+                            </div>
+                        </div>
+                        <Button type="neutral" onClick={setMaxAmount}>
+                            Max
+                        </Button>
+                    </div>
+                </Form.LabeledSet>
+                {isThirdPartyWithdrawal && (
```
sparkplug0025, line 129: Question: Why not allow memo to transfers being sent to itself? Might be helpful for the user history and helps on community visibility

packages/webapp/src/members/transactions.ts
```diff
@@ -0,0 +1,79 @@
...
+    const amount = assetToString(totalWithdrawal, totalWithdrawal.precision);
+
+    const authorization = [
+        {
+            actor: authorizerAccount,
+            permission: "active",
+        },
+    ];
+
+    const sweepAvailableDistributions =
```
sparkplug0025, line 27: question: so, by design, the idea is to sweep everything and only transfer the specified amount, correct?

packages/webapp/.env
sparkplug0025: Yeah, I share your thoughts. I think being more explicit is better indeed. 

About deriving the correct asset from `NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT ="10"` might not be needed as well... This way we are explicitly forcing the configuration to be the correct asset and avoid undesired effects. 

We could check to make sure if the precision and asset from `NEXT_PUBLIC_APP_MINIMUM_DONATION_AMOUNT` is consistent with the `NEXT_PUBLIC_TOKEN_SYMBOL` and `NEXT_PUBLIC_TOKEN_PRECISION` though.

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        availableFunds.precision
+                    )}
+                </span>
+            </Text>
+            <form onSubmit={onPreview} className="space-y-3">
+                <Form.LabeledSet
+                    label={`${tokenConfig.symbol} account name (12 characters)`}
+                    htmlFor="to"
+                >
+                    <Form.Input
```
sparkplug0025, line 88: suggestion: refactor this component to a `Form.ChainAccount` component. We could reuse it in the inductions!

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
```diff
@@ -0,0 +1,150 @@
...
+                        maxLength={12}
+                        pattern="^[a-z,1-5.]{1,12}$"
+                        onInvalid={validateAccountField}
+                        onInput={clearErrorMessages}
+                    />
+                </Form.LabeledSet>
+                <Form.LabeledSet label="Amount to withdraw" htmlFor="amount">
+                    <div className="flex space-x-2">
+                        <div className="relative flex-1">
+                            <Form.Input
```
sparkplug0025, line 103: praise: great component to handle assets! 
suggestion: refactor this component to a `Form.Asset` component! I can totally see it being reused in new features!

sparkplug0025 (approved):
👏 Great job on this PR!

I gave some suggestions which are not blockers for merging this one!

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
brandonfancher: I agree, but the `eosio.token:transfer` action is only called when you transfer funds to a third-party account, and for the time being, no matter what amount you specify for withdrawal/transfer, we sweep all available distribution funds to your personal Eden account (not a partial amount)...so the `fundtransfer` action (which does contain a memo field) might be for a much higher token amount than the `withdraw` and `transfer`. So a memo on the withdrawal itself could end up inaccurate.

We have enhanced fund management on the roadmap, and these are the types of things we'll want to tackle then.

packages/webapp/src/members/transactions.ts
brandonfancher: Exactly. This is the stopgap solution just to get this out there. We'll do a more complete enhanced-funds-management feature in the future.

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
sparkplug0025: oh, I missed that the memo field was only for the `eosio::transfer`! now it makes sense!

packages/webapp/src/members/components/funds-available-cta.tsx
brandonfancher: Good question. I'll double check that.

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
brandonfancher: Done ✅

packages/webapp/src/members/components/withdraw-funds/withdraw-modal-step-form.tsx
brandonfancher: Done ✅

packages/webapp/src/members/components/funds-available-cta.tsx
brandonfancher: Added a flex-wrap. Thanks!

PR #607, opened 2021-11-12
Reset opt-in status when the election is rescheduled.

swatanabe:
This likely also requires front end updates, as it adds multiple versions to some election states and also changes the meaning of the member::election_participation_status field.

### DIFF
--- a/contracts/eden/src/actions/elect.cpp
+++ b/contracts/eden/src/actions/elect.cpp
@@ -32,16 +32,6 @@ namespace eden
 
       members members{get_self()};
       const auto& member = members.get_member(voter);
-      if (participating)
-      {
-         eosio::check(member.election_participation_status() == not_in_election,
-                      "Not currently opted out");
-      }
-      else
-      {
-         eosio::check(member.election_participation_status() != not_in_election,
-                      "Not currently opted in");
-      }
       members.election_opt(member, participating);
    }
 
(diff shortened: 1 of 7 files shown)


contracts/eden/src/elections.cpp
```diff
@@ -450,18 +483,15 @@ namespace eden
       {
          if (iter->status() == member_status::active_member)
          {
-            switch (iter->election_participation_status())
+            printf("%d, %d\n", iter->election_participation_status(),
```
tbfleming, line 486: debugging?

tbfleming:
Since it can automatically bump `election_schedule_version`, this needs an event. Also, eden-micro-chain needs to monitor for the event and clear out members' participating field.

contracts/eden/src/actions/elect.cpp
```diff
@@ -39,7 +39,7 @@ namespace eden
       }
       else
       {
-         eosio::check(member.election_participation_status() == in_election,
+         eosio::check(member.election_participation_status() != not_in_election,
```
tbfleming, line 42: This check and the one above don't match the new definition of whether the user is opted in

swatanabe:
> Since it can automatically bump election_schedule_version, this needs an event

Should `election_event_schedule` be sufficient?

tbfleming:
> Should `election_event_schedule` be sufficient?

yes

tbfleming (approved):

tbfleming:
Future PR: subchain needs to clear opt-in on `election_event_schedule`

PR #609, opened 2021-11-15
Add induction check on donate and fix microchain

sparkplug0025:
- check induction is not expired on inductdonate action
- add a no migration to flag the history about this fix
- add a migrationIndex on the microchain status and captures the event
- check for the migrationIndex before cleaning up expired inductions

### DIFF
--- a/contracts/eden/include/events.hpp
+++ b/contracts/eden/include/events.hpp
@@ -30,6 +30,12 @@ namespace eden
    //
    //    election_event_end
 
+   struct migration_event
+   {
+      uint16_t index;
+   };
+   EOSIO_REFLECT(migration_event, index)
+
    struct election_event_schedule
    {
       eosio::block_timestamp election_time;
@@ -225,7 +231,8 @@ namespace eden
                               distribution_event_return_excess,
                               distribution_event_fund,
                               distribution_event_end,
-                              distribution_event_return>;
+                              distribution_event_return,
+                              migration_event>;
 
    void push_event(const event& e, eosio::name self);
    void send_events(eosio::name self);
--- a/contracts/eden/include/events.hpp
+++ b/contracts/eden/include/events.hpp
@@ -32,7 +32,7 @@ namespace eden
 
    struct migration_event
    {
-      uint16_t index;
+      varuint32 index;
    };
    EOSIO_REFLECT(migration_event, index)
 
--- a/contracts/eden/include/events.hpp
+++ b/contracts/eden/include/events.hpp
@@ -5,6 +5,7 @@
 #include <eosio/name.hpp>
 #include <eosio/reflection.hpp>
 #include <eosio/time.hpp>
+#include <eosio/varint.hpp>
 #include <variant>
 #include <vector>
 
@@ -32,7 +33,7 @@ namespace eden
 
    struct migration_event
    {
-      varuint32 index;
+      eosio::varuint32 index;
    };
    EOSIO_REFLECT(migration_event, index)
 
--- a/contracts/eden/src/eden-micro-chain.cpp
+++ b/contracts/eden/src/eden-micro-chain.cpp
@@ -12,6 +12,7 @@
 #include <eosio/ship_protocol.hpp>
 #include <eosio/to_bin.hpp>
 #include <events.hpp>
+#include <migrations.hpp>
 
 using namespace eosio::literals;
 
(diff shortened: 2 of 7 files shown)


contracts/eden/src/migrations.cpp
```diff
@@ -7,15 +8,16 @@ namespace eden
       eosio::check(migration_sing.get().index() == new_value.index(),
                    "Cannot change current migration");
       migration_sing.set(new_value, contract);
+      push_event(migration_event{static_cast<uint16_t>(new_value.index())}, contract);
```
swatanabe, line 11: This is the wrong place to emit the event.  This function is not used by the `migrate` action; it's used by other actions that affect the state of the migration.

contracts/eden/include/events.hpp
```diff
@@ -30,6 +30,12 @@ namespace eden
    //
    //    election_event_end
 
+   struct migration_event
+   {
+      uint16_t index;
```
swatanabe, line 35: Why uint16_t?  variant stores the index as a varint, and we're unlikely to exceed a single byte in the near future.

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1853,19 +1860,12 @@ void call(void (*f)(const action_context&, Args...),
...
+void remove_expired_inductions(const eosio::time_point& block_time, const status& status)
 {
-   auto& idx = db.status.get<by_id>();
-   if (idx.size() < 1)
-      return;  // skip if genesis is not complete
-
-   const auto& status = get_status();
-   if (!status.status.active)
-      return;  // skip if not active
+   if (status.migrationIndex < 5)
```
swatanabe, line 1865: Can we please not use a magic number?  You can derive the correct index from the variant, as in `migrations::is_completed`

swatanabe (changes requested):

contracts/eden/src/migrations.cpp
sparkplug0025: Could you point me to the correct place to emit the event please? What about the next one, inside `init()`?

contracts/eden/include/events.hpp
sparkplug0025: Ah oops my bad changing to `uint8_t`

tbfleming: `varuint32`. This is 1 byte for the near future, but expands as needed.

contracts/eden/src/migrations.cpp
tbfleming: `init` happens at `genesis`. `migrate_some`.

swatanabe: It should be emitted any time the type held in the variant changes.  That means `init` and `migrate_some`.  `set` is not needed as it explicitly forbids changing the type held by the variant.

swatanabe: > `init` happens at `genesis`. `migrate_some`.

emitting the event from `init` allows the microchain to know what version of the contract was deployed at genesis.

contracts/eden/src/migrations.cpp
```diff
@@ -43,6 +42,7 @@ namespace eden
              state);
       }
       migration_sing.set(state, contract);
+      push_event(migration_event{static_cast<varuint32>(state.index())}, contract);
```
swatanabe, line 45: This can cause the contract to send multiple events with the same index.  Is that what we want?

contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -218,6 +219,12 @@ struct status
    uint16_t electionThreshold = 0;
    uint16_t numElectionParticipants = 0;
    uint16_t migrationIndex = 0;
+
+   template <typename T>
+   bool isMigrationCompleted() const
+   {
+      return migrationIndex >= boost::mp11::mp_find<eden::migration_variant, T>::value;
```
swatanabe, line 226: The of this function is incorrect.  It returns true if the migration has started, not if it has completed.

contracts/eden/src/migrations.cpp
sparkplug0025: Nice catch, it's not, I'm fixing it to be triggered only when the event actually finishes.

contracts/eden/src/eden-micro-chain.cpp
sparkplug0025: With my above change I think it's fixed, can you please check it again?

swatanabe (approved):

PR #610, opened 2021-11-15
Break election components apart

brandonfancher:
This componentizes and organizes the election UI. After this PR, we can refactor this to depend on subchain queries. And I'm sure we'll find more opportunities for refactoring as we do that.


contracts/eden/tests/run-start-election.cpp
```diff
@@ -0,0 +1,16 @@
...
+TEST_CASE("Setup Eden chain with start of full election")
+{
+   nodeos_runner r("chain-start-election");
+
+   r.tester.genesis();
+   r.tester.run_election(true, 10000, true);
+   r.checkpoint("small_election");
+   r.tester.induct_n(100);
+   r.checkpoint("inductions");
+   r.tester.eden_gm.act<actions::electsettime>(s2t("2021-11-16T18:45:00.000"));
```
brandonfancher, line 12: I'd like for this to always pull the current datetime so that the election is active after this runs...but haven't figured that out yet. That can be done in a subsequent PR.

thesalva101: nice touch!

sparkplug0025: nice touch!

contracts/eden/tests/run-start-election.cpp
```diff
@@ -0,0 +1,16 @@
...
+   r.tester.start_election(true, 10000);
```
sparkplug0025, line 13: suggestion: we can centralize the elections scenario in the current `run-full-election.wasm` file and use checkpoints to stop the runner where we want. For instance, as opposed of creating this file, just add this `start_election` to the bottom of the full-election one with a checkpoint:

```cpp
   r.tester.genesis();
   r.tester.run_election(true, 10000, true);
   r.checkpoint("small_election");
   r.tester.induct_n(100);
   r.checkpoint("inductions");
   r.tester.eden_gm.act<actions::electsettime>(s2t("2021-11-16T18:45:00.000"));
   r.checkpoint("full_election1");
   r.tester.start_election(true, 10000);
```

If someone wants to avoid the start of the new election they can just do `cltester run-full-elections.wasm -c full_election1`

I also think since we are using checkpoints we should rename the file from `run-full-elections` to `run-elections`, then it's more generic about elections scenarios....

brandonfancher: Oh, cool! That’s awesome!

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round/round-info/meeting-link/meeting-link.tsx
```diff
@@ -18,7 +18,7 @@ import { useEncryptionPassword, usePasswordModal } from "encryption";
 
 import MeetingButtons from "./meeting-buttons";
 import MeetingLinkModal from "./meeting-link-modal";
-import { election as electionEnvVars } from "config";
+import { election as electionConstants } from "config";
```
sparkplug0025, line 21: nit: not big deal, but maybe `electionConfig`? `constants` appear that these values are set in stone, whereas we actually allow the community to set their own election params...

sparkplug0025 (approved):
LGTM! With the checkpoint suggestion you can remove the new files and cmake changes :)

packages/webapp/src/elections/components/ongoing-election-components/ongoing-round/round-info/meeting-link/meeting-link.tsx
brandonfancher: For now, I'll switch back to `electionEnvVars`. It's not great, but `electionConfig` is a variable already declared in this file. This will likely change as the refactor continues.

PR #613, opened 2021-11-18
clsdk enhancements and documentation

tbfleming:
* New `activate_feature` contract provides an easy way to activate features in test cases
* New `set_parameters` contract provides an easy way to modify blockchain parameters in test cases
* `test_chain` 's constructor now lets tests control the maximum state database size
* `test_chain::start_block(std::string_view time)` and `test_chain::start_block(time_point tp)` can move the chain to specific times.
* `test_chain::set_abi` loads an .abi file from disk and installs it.


contracts/CMakeLists.txt
```diff
@@ -2,3 +2,5 @@ add_subdirectory(eden)
 add_subdirectory(token)
 add_subdirectory(boot)
 add_subdirectory(bios)
+add_subdirectory(activate_feature)
+add_subdirectory(set_parameters)
```
swatanabe, line 6: Why use new contracts rather than boot/bios which serve a similar purpose.  Switching out the system contract behind the scenes seems like a good way to cause surprises.

programs/cltester/main.cpp
```diff
@@ -1086,7 +1087,13 @@ struct callbacks
 
    uint32_t tester_create_chain(span<const char> snapshot)
    {
-      state.chains.push_back(std::make_unique<test_chain>(state, span_str(snapshot).c_str()));
+      return tester_create_chain2(snapshot, 1024 * 1024 * 1024);
```
swatanabe, line 1090: I'd prefer `chain::config::default_state_size` over hard-coding.

contracts/CMakeLists.txt
tbfleming: `boot` has a fixed config, suitable for Eden testing. I wanted a more modular system that doesn't leave a bios contract installed, unless the user wants a bios contract.

swatanabe: The whole point of the bios contract is to allow access to configuration like `set_parameters`.  The only real reason not to want the bios contract is if you're installing your own system contract.  The point of the boot contract (at least, the original point--we've shoved more into in since) was to make the bios loadable.

tbfleming: Over time, the bios contract would depend on more HFs being activated, making it an all-or-nothing, unless clsdk builds the bios contract while building the user's test suite.

swatanabe: That's true, but
- Not all protocol features need configuration
- Very few contract developers will care if they get a few extra protocol features.
- `preactivate_feature` is must-have which the tester activates unconditionally and `set_blockchain_parameters_packed` is baseline.

If we do create separate contracts I would strongly prefer to create additional accounts for them instead of blanking out eosio every time the user runs some configuration.

tbfleming: Additional accounts adds the `set_privileged` headache. We can clean up `boot` and `bios` in the future, giving test authors a choice between the two approaches.

swatanabe: Why would users want a choice?  They won't understand what they're choosing.  They might even try to use both in the same test, which would be bad.

I think the side-effect of clearing code on eosio when setting configuration is unacceptable.  I'll take pretty much any other approach over that.

bytemaster: I agree with Steven

swatanabe: A few ideas:
- Reset code to its original value.  Side-effect: increments code sequence.  I think this is still ugly, but I can live with it.
- Fail if eosio has a contract.  This doesn't prevent errors, but it will tell users what they did wrong.
- Bios + protocol feature specific contracts.  Precondition: eosio must hold a compatible contract, with setpriv at a minimum.  Side-effect: extra accounts created.
- bios provides all functionality.  Some protocol features must be activated during setup, whether users want them or not.
- bios provides base functionality and handles protocol feature specific behavior either by modifying itself or deploying additional contracts on demand.  The implementation is painful, but the user-visible behavior is easy to describe.

doc/clsdk/src/contract/debug/vscode.md
```diff
@@ -0,0 +1,98 @@
...
+* continue
+* call stack
+
+The following are not available
+* examining variables
+* examining memory
+
+## Corrupted database recovery
+
+The debugger can cause nodeos to corrupt its database. There are 2 options to recover from the corruption:
```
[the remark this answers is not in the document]
tbfleming: That's the primary way. It's also possible to redirect signal handling, move the IP, or do other things which mess up execution.

swatanabe (approved):

PR #614, opened 2021-11-18
Get UI working with ephemeral chain

brandonfancher:
Refactoring areas of UI dependence on AtomicHub APIs for member data to instead use subchain. For example, the account profile component in the nav bars, My Delegation page, and ongoing election UI was broken when running on the ephemeral chain. This fixes that.

While I need to finish testing a full election before I know everything is working properly, this does represent important progress. So I'll go ahead and submit this for review.

Once UI-working-with-ephemeral-chain baseline is met, we can get into deeper refactoring.

### DIFF
--- a/packages/webapp/src/_app/hooks/queries.ts
+++ b/packages/webapp/src/_app/hooks/queries.ts
@@ -8,6 +8,7 @@ import {
     getMembersStats,
     MemberData,
     MemberStats,
+    useMembersByAccountNames,
 } from "members";
 import { getCommunityGlobals, getTokenBalanceForAccount } from "_app/api";
 import {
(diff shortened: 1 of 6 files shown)


brandonfancher:
By the way, I LOVE ephemeral chain. So much easier (once you get the hang of it). I even added it as a custom network in bloks.io and Anchor so I can use those tools too. 💯

sparkplug0025 (approved):
praise: amazing to see things getting simpler!

> By the way, I LOVE ephemeral chain. So much easier (once you get the hang of it). I even added it as a custom network in bloks.io and Anchor so I can use those tools too. 💯

We need to productize it correctly and sell the idea in a nice way: no-brainer, small demoable flow in a landing page/hotsite!

PR #615, opened 2021-11-19
Add Eden `execsession` server signature to box

sparkplug0025:
- Adds the `/v1/sessions/sign` api to the eden box
- Implements a TaposManager based on `democlsdk/serverpay`
- Signs the eden contract `execsession` action with a server key 
  - The actions are not being whitelisted or checked yet, awaiting for interface definitions...


brandonfancher (approved):
Looks good to me! Probably worth getting @tbfleming's eyes on this too.

packages/box/src/handlers/sessions.ts
```diff
@@ -0,0 +1,87 @@
...
+    const trx = prepareExecSessionTrx(sessionSignRequest);
+    const signedTrxResult = await signTrx(trx);
+    logger.info(
+        `signed [execsession] for account:${sessionSignRequest.edenAccount}
+        sequence:${sessionSignRequest.sequence}`
+    );
+    return res.json(signedTrxResult);
+};
+
+const prepareExecSessionTrx = (sessionSignRequest: SessionSignRequest) => {
```
tbfleming, line 54: This is missing the no-op action which pays for the transaction

sparkplug0025: But the transaction is a single `execsession` being signed by the server, no? That being the case, do we need a noop serverpay action?

sparkplug0025: @tbfleming I addressed this in a new commit, can you please give another round of reviews?

tbfleming (approved):

PR #620, opened 2021-11-24
Separate out NFT data from Member data

brandonfancher:
## Description

This is a continuation of the gradual refactor of the frontend. 

Data in the `MemberData` interface comes from AtomicAssets. It _is_ the member's NFT information. `Member` data on the other hand now comes from the subchain. In the future, a member's _current_ profile information may differ from the (older) information in their NFT.

For this reason, and for the sake of clarity--especially as we refactor--I have renamed `MemberData` to `MemberNFT`. The real data for a member will now be found in the `Member` interface/model.

To restate the rule from now on: If we want a member's profile or current Eden membership status/data, we should get that from the subchain/box (via the `Member` interface). If we want a member's NFT information (including profile information as minted in that NFT), we should get that from the AtomicHub APIs (via the `MemberNFT` interface).

This will also make it easier for us to remove NFTs from the platform, if we so choose.

## In this PR, I also...

- [x] adjust member chips to accept either `Member` or `MemberNFT`, so I can...
- [x] make the main members list page use `Member` and box queries,
- [x] organize/move NFT interfaces into the NFT domain

## Future work
- continue refactoring remaining UI (_e.g._, elections, inductions) to use box queries and the new interfaces

### DIFF
--- a/packages/webapp/src/_app/hooks/box-queries.ts
+++ b/packages/webapp/src/_app/hooks/box-queries.ts
@@ -7,9 +7,9 @@ import dayjs from "dayjs";
 import { assetFromString } from "_app";
 import {
     formatMembersQueryNodeAsMemberNFT,
-    MemberNFT,
     MEMBER_DATA_FRAGMENT,
 } from "members";
+import { MemberNFT } from "nfts/interfaces";
 
 export interface ElectionStatusQuery {
     status: {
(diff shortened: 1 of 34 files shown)


brandonfancher:
> Then I was thinking that the MemberNFT was not being used in the proper places (eg: voting rounds should not care about nft data, right?). But by reading some code comments regarding future refactors it seems like it will come later....

Correct. Just haven't gotten to refactoring those areas yet.

PR #625, opened 2021-12-09
Add Signing session transactions example page

sparkplug0025:
A page with examples of how a session key can be created and how it can be used to sign a transaction was created in this PR. For that, it was needed to create the frontend functionality to create and store the session key (using IndexedDb) and the glue between the frontend and eden box.

To test it:
1. Spin up an ephemeral chain
1. Go to https://localhost:3000/sessions
1. Open the inspector to observe the console
1. Click on create session key
1. Click on induct Ahab to see that we are using the created sessionkey
1. You can see the created induction on the membership page

Next steps:
- Work on getting the proper sequence number dynamically (persist it on the microchain and get it from the useCurrentMember query)
- Work on the UI flows to create session keys, detected when it's created and use it on allowed transactions in the UI
- Work on the re-generate session key flow when it expires (it has a limit of 90 days in the contract and in the current creation we are actually setting it to 30 days)
- Migrate from eosjs to SubtleCrypto

In the meanwhile:
- We could go ahead and create an eden payer account in Wax testnet and EOS mainnet and set the proper env configs for the eden boxes in these environments.


packages/webapp/src/_app/eos/sessions.ts
```diff
@@ -0,0 +1,163 @@
...
+export const signSessionTransaction = async (
+    authorizerAccount: string,
+    actions: any[]
+) => {
+    const sessionKey = await sessionKeysStorage.getKey();
+    if (!sessionKey) {
+        throw new Error("Session key is not present");
+    }
+
+    const sequence = 1; // TODO: get sequence dynamically
```
sparkplug0025, line 92: todo in a next pr. for now, it will only work once per each ephemeral chain. (increase it manually if you want to run further tests in the same chain)

packages/webapp/src/pages/sessions.tsx
```diff
@@ -0,0 +1,97 @@
+import { hexToUint8Array } from "eosjs/dist/eosjs-serialize";
```
sparkplug0025, line 1: this is an example page as a reference of how to actually create and sign with sessions.

packages/webapp/src/_app/eos/sessions.ts
tbfleming: Instead of getting it dynamically, you could store the sequence number in SessionKeyStorage. Bump it up by 1 with each use. If the user clicks too fast and starts getting sequence errors, maybe provide a button to refresh it from the chain, or delete the session key and start over (must be with a newly-generated key to prevent a replay attack).

tbfleming (approved):

PR #647, opened 2021-12-14
Enable LTO for wasm build

swatanabe:
Reduces the number of function definitions in the contract to fit in the limits.


tbfleming:
I don't want to force lto into the sdk. Setting the option directly on eden works well enough for now. We might later add -lto targets to the sdk.

swatanabe:
Why not?  LTO objects can still be linked normally without LTO.

tbfleming:
I suspect wasm-ld has a special rule that automatically triggers lto when there's an lto object. Native builds fail when linking lto objects into an executable without giving the linker the lto flag.

tbfleming (approved):

PR #654, opened 2022-01-04
WIP add session signature with SubtleCrypto

sparkplug0025:


packages/webapp/src/_app/eos/sessions.ts
```diff
@@ -194,7 +223,152 @@ const serializeSignatureAuth = async (
...
+        //     await crypto.subtle.digest("SHA-256", data)
+        // );
+
+        // const ecSignature = {
+        //     r: signatureBytes.slice(0, 32),
+        //     s: signatureBytes.slice(32, 64),
+        // };
+        // console.info(ecSignature, eosPubKey, pubKey, hash, signatureBytes);
+        // const recid = EC.getKeyRecoveryParam(hash, ecSignature, pubKey);
+        // const recidByte = 27 + 4 + recid;  <<<--- still presenting failures
```
sparkplug0025, line 262: this recidByte still fails sometimes with the following error:
`invalid high s-value encountered in r1 signature`

packages/webapp/src/_app/eos/sessions.ts
```diff
@@ -194,7 +223,152 @@ const serializeSignatureAuth = async (
...
+        let recoveredKey = signatureEcObj.recover(data, true).toString();
+
+        console.info(
+            "recoveredKey >>>",
+            recoveredKey.toString(),
+            sessionPubKeyString
+        );
+
+        if (recoveredKey.toString() === sessionPubKeyString) {
+            return signatureStr;
```
sparkplug0025, line 291: still gives me the same error above, even when I can confirm the recovered public key from elliptic lib matches.

PR #667, opened 2022-02-02
cltester: eosio.system example

tbfleming:
Also improves `set_code` and `set_abi` interface


tbfleming:
moving PR to new repo

PR #671, opened 2022-02-23
microchain: support rename

tbfleming:


contracts/eden/src/eden-micro-chain.cpp
```diff
@@ -1662,6 +1662,94 @@ void resign(eosio::name account)
...
+
+   db.status.modify(get_status(), [&](auto& status) { update_vec(status.status.initialMembers); });
+
+   if (auto* obj = get_ptr<by_pk>(db.balances, old_account))
+      db.balances.modify(*obj, [&](auto& obj) { obj.account = new_account; });
+
+   for (auto& obj : db.balance_history)
+      db.balance_history.modify(obj, [&](auto& obj) {
+         update(obj.account);
+         update(obj.other_account);
```
swatanabe, line 1683: It's a bit unfortunate that other_account forces us to iterate over the whole table.

swatanabe (approved):

PR #674, opened 2022-03-21
Front end bug fixes

brandonfancher:
This PR fixes the following issues:

- If user signs in as a non-member EOS account, they cannot sign out without clearing local storage. This was a regression introduced in a prior refactor.
- A member's NFT collection was not displaying; it was trying to fetch more items than allowed by the AtomicMarket API (a limit which they randomly began enforcing). We fixed this previously for a member's list of collectORS, but not for their collectION.
- Console was complaining that a React component key was missing from a component on the Treasury page.

Note: Each commit maps 1:1 with a bug above, so it's likely easiest to review this commit-by-commit.

### DIFF
--- a/packages/webapp/src/_app/ui/nav-profile.tsx
+++ b/packages/webapp/src/_app/ui/nav-profile.tsx
@@ -16,10 +16,10 @@ interface Props {
 }
 
 export const NavProfile = ({ location }: Props) => {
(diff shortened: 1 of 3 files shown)


GreenBusDriver (approved):
LGTM; just the e2e tests seem to be failing. I assume you're looking at that as I type

cool-ant (approved):

PR #679, opened 2022-03-22
remove_expired_inductions is still borked

tbfleming:


brandonfancher (approved):
Are there any consequences for not having this enabled?

tbfleming:
It's been effectively disabled this whole time because of the rollout failure. It just means expired inductions may appear in the UI as still valid.

PR #694, opened 2022-09-20
Values for mock election jungle4

fagomezra:


fagomezra:
this PR it's for edeni:dev.

